PR #68, opened 2026-06-11
fix(dagster): share the HelmRepository via a singleton so multiple instances don't collide

yehudacohen:
## Problem

The Dagster bootstrap created its Flux `HelmRepository` **inline** in the RGD, so each KRO instance's ApplySet tried to *own* `flux-system/dagster` exclusively. Deploying a **second** instance of the same RGD (e.g. a dev + prod pair) then failed to reconcile:

```
resource belongs to a different ApplySet: flux-system/dagster
  (source.toolkit.fluxcd.io/v1, Kind=HelmRepository) ... cannot reassign to ApplySet "..."
```

The HelmRelease and pods came up fine — only KRO's per-instance ownership of the shared repository conflicted.

## Fix

The official Dagster chart repository is **one cluster-level Flux source** shared by every instance, so model it as a shared singleton rather than a per-instance owned resource — the same pattern the webapp factory uses for its cluster-scoped operators.

- Add a `dagsterHelmRepositoryBootstrap` composition (kind `DagsterHelmRepository`) that owns the `HelmRepository` and exposes a `ready` status.
- `dagsterBootstrap` consumes it via `singleton(...)` with a **fixed cluster-wide identity** (official `dagster` / `flux-system` / chart URL). Each instance now owns only its `Namespace` + `HelmRelease` and references the shared repository by the official `sourceRef` defaults; the repository lives outside any instance's ApplySet.
- Because the shared repo's identity is fixed, the per-instance `repositoryName` / `repositoryNamespace` **bootstrap-spec** fields (which never made sense for a shared source) are removed. ⚠️ Breaking schema change for anyone who set them. The low-level `dagsterHelmRelease` resource factory keeps its own repository options.

## Verification

Deployed live on a cluster: a **dev + prod pair** of `DagsterBootstrap` instances both reconcile to `ACTIVE`/`Ready` sharing one `DagsterHelmRepository` singleton (deployed once in `typekro-singletons`), with **zero** "different ApplySet" errors.

- Full suite: **4386 pass / 0 fail**
- typecheck, lint, build, madge (no circular), docs checks: all clean

> Note (CRD migration): because this removes two CRD properties, upgrading an existing install requires deleting the old `dagsterbootstraps.kro.run` CRD first — KRO refuses breaking in-place CRD schema changes. Fresh installs are unaffected.

Independent of #67 (no file overlap); both target `master`.

### DIFF
--- a/docs/api/dagster/index.md
+++ b/docs/api/dagster/index.md
@@ -21,11 +21,18 @@ import {
 
 ## What Gets Created
 
-`dagsterBootstrap` creates three TypeKro-owned resources:
+Each `dagsterBootstrap` instance owns:
 
 - A `Namespace` for the Dagster deployment namespace.
-- A Flux `HelmRepository` for `https://dagster-io.github.io/helm`.
 - A Flux `HelmRelease` for chart `dagster`, default version `1.13.8`.
+
+The Flux `HelmRepository` for `https://dagster-io.github.io/helm` is a **shared
+cluster-level singleton** (`DagsterHelmRepository`), deployed once and referenced by every
+Dagster instance. Modelling it per-instance would make each instance's KRO ApplySet try to
+own the same `flux-system/dagster` HelmRepository exclusively, so a second instance (e.g. a
+dev + prod pair) would fail to reconcile. The singleton owner lives in the `typekro-singletons`
+namespace; the factory's `deploy()` creates it automatically. See
+[Singletons & GitOps](#singletons-and-gitops) for the GitOps (`toYaml`) workflow.
 
 The bootstrap status is derived from owned Flux resources only. Component booleans such as
 `webserver`, `daemon`, and `userDeployments` mean the owning HelmRelease is ready, not that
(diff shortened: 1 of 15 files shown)


yehudacohen:
Addressed the GitOps finding (singleton owner omitted from `toYaml()`) in 58a7936.

**Root cause:** `toYaml()` was singleton-blind. `dagsterBootstrap.toYaml()` emitted only the consuming RGD (externalRef to the `DagsterHelmRepository` singleton); the owner RGD + owner instance were created only by the factory's imperative `deploy()`. A GitOps apply of the generated YAML left the externalRef dangling → broken deploy. This affected every singleton-using composition (webapp's cnpg/valkey too) — dagster is just the first shipped via a GitOps-documented path.

**Fix:** `toYaml()` is now singleton-aware (new `core/deployment/singleton-gitops.ts`):
- `composition/factory.toYaml()` (no spec) emits the singleton owner RGD(s) **deps-first** before the consuming RGD.
- `factory.toYaml(spec)` emits the shared owner instance(s) before the consuming instance — built to match `deploy()` exactly (derived name, registry namespace, `typekro.io/singleton-spec-fingerprint` annotation), so the externalRef resolves *and* a later imperative `deploy()` passes its drift check.
- Isolation preserved: owners are separate documents, never injected into the consuming graph.
- Dagster docs corrected ("What Gets Created" + new "Singletons and GitOps" section).

**Verified live:** pure `kubectl apply` of `toYaml()` output (no factory) for a dev + prod pair → both `ACTIVE/Ready`, sharing one HelmRepository singleton, **zero ApplySet collisions**. Full suite 4377/0; typecheck/lint/build/madge/docs clean.

PR #76, opened 2026-06-14
feat(caddy): config-driven Caddy reverse-proxy integration

yehudacohen:
## Summary

Adds **`typekro/caddy`** — a *config-driven* Caddy reverse proxy. It's **not** the Caddy ingress-controller and **not** a Helm bootstrap: the composition runs the official `caddy` image with a Caddyfile you supply and emits **ConfigMap + Deployment + Service + PVC**.

Caddy's built-in **`tls internal`** issues certificates from a local CA — so there's **no cert-manager, no etcd, no ACME**. The PVC persists `/data` so the CA root is stable across restarts.

## What's included

- **`caddyIngress`** composition (kro + direct), schema-first types, status derived from direct `Deployment`-proxy comparisons (`ready` / `phase` / `version`)
- **`renderCaddyfile(routes, { tls })`** helper — builds the Caddyfile string from concrete `host → upstream` routes. Kept *out* of the composition so the composition is a pure string passthrough (KRO-safe: a structured `routes` array can't be `.map()`-ed into a string in KRO mode)
- unit tests (`renderCaddyfile`, composition RGD/schema/factory modes, exports) + a live integration test
- docs page (`docs/api/caddy/`) + sidebar entry + `package.json` `./caddy` export

```ts
import { caddyIngress, renderCaddyfile } from 'typekro/caddy';

const caddyfile = renderCaddyfile([
  { host: 'dagster-dev.acme.internal', upstream: 'dagster.dagster-platform-dev.svc:80' },
]);
await caddyIngress.factory('direct', { namespace: 'caddy-system' })
  .deploy({ name: 'caddy', caddyfile });
```

## Design note — a new "config-driven workload" pattern

The existing factories (dagster, apisix, cert-manager) are **Helm bootstraps**. Caddy here is a different shape: render-config + raw `Deployment`/`ConfigMap`/`Service`/`PVC` + direct-proxy status (no `helm-values-mapper`, no `Cel.expr`-from-conditions). Two sharp edges worth folding into the integration skill, which is currently Helm-centric:
- direct-comparison status requires the **const name to match the resource `id`** (`const caddyDeployment = deployment({ id: 'caddyDeployment' })`), otherwise the status CEL references an unresolvable variable
- arktype validation failures are `ArkErrors`, not `Error` — tests assert `instanceof type.errors`

## Live validation (EKS)

Direct deploy reached `status.ready === true` and the pod logs confirm the core claim:

```
certificate obtained successfully  identifier=whoami.caddy-e2e.internal  issuer=local
certificate installed properly in linux trusts
serving initial configuration
```

`tls internal` minted a cert from its **local CA** for the configured host and Caddy served HTTPS; `deploy` / `deleteInstance` were clean. Default image pinned to the current stable **`caddy:2.11.2`**.

## Test plan

- `bun run typecheck` ✅ · `biome check` ✅
- `bun run test test/factories/caddy/` ✅ (render per TLS mode, RGD/schema/factory, exports)
- `bun test test/integration/caddy/` ✅ live on EKS (deploy → ready, RGD + factory modes)

## Follow-ups (not blocking)

- Full **KRO-mode live deploy** (direct deploy proven live; kro proven via factory + RGD generation; integration isn't dependency-managing, so kro applies the same resources)
- A `curl`-with-CA proxy-to-backend e2e (cert + HTTPS serving already proven in the logs; `reverse_proxy` is Caddy core)


yehudacohen:
Pushed `84885cd` addressing all three review findings:

1. **Hook timeout broke `bun run typecheck`** — the integration `afterAll` passed a second timeout arg to the one-arg `deleteNamespaceAndWait` (TS2554). Switched to the non-waiting `deleteNamespaceIfExists` (initiates teardown and returns under bun's 5s default hook timeout), so no per-hook timeout arg is needed. Full `bun run typecheck` is clean.

2. **KRO readiness ignored `replicaCount`** — `readyReplicas >= replicas` baked the JS `replicaCount ?? 1` literal into the CEL. Now compares against the Deployment's reflected desired count: `status.readyReplicas >= status.replicas`. Both operands are `.status` fields, so it resolves in **both** kro CEL and direct-mode status hydration (`spec.replicas` resolves only in kro — caught that the hard way: a `spec.replicas` version passed unit tests but reported `ready: false` live in direct mode).

3. **Default image dereferenced an optional version** — `${image}:${string(version)}` yielded `caddy:` when `version` was unset in kro. `image` is now a single full ref (`caddy:2.11.2`) applied as one `?? DEFAULT_CADDY_IMAGE` field → compiles to `has(spec.image) ? spec.image : "caddy:2.11.2"`.

Also dropped `status.phase`: the `ready ? 'Ready' : 'Installing'` ternary serialized to malformed CEL (`caddyDeployment.schema.spec.replicas`) — a separate analyzer bug; `ready` carries the signal.

Added regression tests asserting the readiness + image CEL. Re-validated live on sela-eks: direct deploy → `status.ready === true` (2 pass, 0 fail).

yehudacohen:
Folded in a proper fix for the direct-vs-kro split I hit while fixing review item #2.

**Root cause** (`d9e3ec3`/`bda5135`): direct-mode status hydration silently dropped `.spec`. `buildLiveStatusMap` fetches the full live object via `k8sApi.read()` but stored only `.status`, and both proxy sites gated live resolution on `basePath === 'status'`. So a `spec.*` ref (e.g. the canonical `readyReplicas >= spec.replicas` readiness idiom) stayed an unresolved proxy → `NaN` comparison → `false` — while in kro CEL the same ref resolves. A composition correct in kro was quietly wrong in direct.

**Fix:** stash the live `.spec` (already fetched, just discarded) on the same map entry under a new `LIVE_SPEC_KEY` Symbol, and teach `create-resource.ts` + `imperative.ts` to resolve `spec.*` from it during re-execution. A Symbol keeps spec out of `Object.keys`/JSON/status iteration and can't collide with a real status field. Reading live spec is also *more* correct than the desired literal — it reflects cluster defaulting / mutating webhooks.

With that in core, Caddy readiness goes back to `status.readyReplicas >= spec.replicas` (the true desired count; avoids the t=0 transient where `status.replicas` is briefly 0).

**Verification:** new core unit tests (spec resolves from the stash, drives a correct readiness verdict, falls back to a ref when unstashed); full unit suite 4368/0; Caddy live e2e on sela-eks 2 pass/0 fail with `spec.replicas` → `status.ready === true`.

yehudacohen:
Fixed two soundness bugs the readiness work surfaced (`661f3b5`):

1. **Readiness could go true before the desired pod exists.** Comparing to `status.replicas` (the reflected count) is `0` at t=0 → `0 >= 0` true. Now compares to `spec.replicas` (a concrete ≥1), so `ready` only trips once the one pod is actually ready.

2. **`replicaCount` accepted multi-pod configs that share one RWO PVC.** The `tls internal` CA lives in the `ReadWriteOnce` `/data` volume — multiple pods can't co-mount it, and each would mint a *different* CA (clients see mismatched certs). This is a single-replica design.

Changes:
- Removed `replicaCount` from the schema, and set `'+': 'reject'` so unsupported keys (replicaCount, typos) fail **loudly** instead of being silently dropped.
- Pinned the Deployment to `replicas: 1` with `strategy: Recreate` — a `RollingUpdate` would surge a second pod that can't co-mount the RWO PVC and wedge the rollout.
- Docs note the single-replica design + the HA path (RWX storage or externalized CA); tests assert the pin, the strategy, and the key rejection.

Re-validated live on sela-eks: deploy → `ready=1/1`, `status.ready === true`, 2 pass/0 fail.

PR #85, opened 2026-06-29
fix: KRO-invalid CEL for optional scalar fields in runtime values-merge + alchemy RGD timeout ignored [0.19.1]

yehudacohen:
## 0.19.1 — two fixes for a KRO converge that 0.19.0 broke

Adopting 0.19.0 in a downstream (a dagster platform converge) **failed**: the generated RGD was rejected by KRO, and the deploy then aborted on a too-short timeout. Two distinct bugs, both fixed here.

### 1. KRO-invalid CEL for optional **scalar** fields in the runtime values-merge

When a `values` block contains a CEL/ref (forcing the runtime map-merge, `needsRuntimeMapMerge`), each optional overlay field was emitted as a value inside a `.merge({...})` map literal with an `omit()` fallback:
```
.merge({ "nameOverride": has(spec.nameOverride) ? spec.nameOverride : omit(), ... })
```
KRO types `omit()` as `map(string, dyn)`, so for a **scalar** field the ternary is `bool ? <scalar> : map(string,dyn)` — both branches must share a type, so cel-go rejects the RGD:
```
GraphAccepted=False  InvalidResourceGraph
ERROR: <input>:1:20268: found no matching overload for '_?_:_' applied to '(bool, string, map(string, dyn))'
```
**Latent since v0.11** — static values maps never take this path; it only triggers when something puts a CEL/ref in the values (e.g. a conditional default). Tests missed it because they assert CEL *strings*, never compile them against KRO.

**Fix:** optional bare schema refs are pulled out of the inline map and emitted as a type-safe conditional single-key merge — `.merge(has(spec.X) ? {"X": spec.X} : {})` — both branches maps, correct for fields of any type, a no-op when absent (true omit semantics). Static maps and the field-level `has(x) ? x : omit()` form are untouched.

### 2. Alchemy RGD deploy ignored the configured `timeout` (hardcoded 60s)

The alchemy RGD-declaration path hardcoded `DEFAULT_RGD_TIMEOUT` (60s) for the readiness wait, ignoring `factoryOptions.timeout` (the non-alchemy paths already honor it). With #1 fixed, the RGD is accepted and a Helm workload rollout (e.g. a daemon restart from a values change) can legitimately take >60s to reach ready — which then false-failed with `AbortError: Delay aborted`. Now honors `factoryOptions.timeout || DEFAULT_RGD_TIMEOUT`.

## Tests / gates
- Regression test (serialization): forces the runtime merge with optional scalar overlay fields; asserts the type-safe single-key-merge form and **no** scalar-vs-`omit()` ternary. Revert-test confirmed it fails on 0.19.0 and passes with the fix.
- Note: KRO compile-correctness isn't unit-verifiable here (no cel-go in the repo); the structural assertion is the guard, and the fix uses only map/map ternaries (basic, unambiguously valid CEL).
- `tsc --noEmit` clean; unit suite green (one unrelated machine-load-sensitive perf microbenchmark flakes intermittently).
- Version bumped to `0.19.1`; CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.19.1] - 2026-06-29
+
+### Fixed
+
+- Fixed KRO-invalid CEL emitted for **optional scalar** spec fields in the runtime
+  values-merge path. When a `values` block contained a CEL/ref (forcing the runtime
+  map-merge), each optional overlay field was emitted as
+  `.merge({ "X": has(spec.X) ? spec.X : omit() })`. KRO types `omit()` as
+  `map(string, dyn)`, so for a scalar field the ternary `bool ? <scalar> : map(string, dyn)`
+  failed to type-check in cel-go with `GraphAccepted=False` /
+  `found no matching overload for '_?_:_' applied to '(bool, string, map(string, dyn))'`.
+  Optional refs in a runtime merge now emit a type-safe conditional single-key merge
+  `.merge(has(spec.X) ? {"X": spec.X} : {})` (both branches maps) — correct for fields
+  of any type, and a no-op when the field is absent. Static values maps and the
+  field-level `has(x) ? x : omit()` form are unchanged.
+
 ## [0.18.0] - 2026-06-26
 
 ### Added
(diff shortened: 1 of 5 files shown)


yehudacohen:
Follow-up (`65211e8`) addressing review feedback: the conditional single-key merge was emitting `optionalRefMergeGuard`'s **inner** schema path as the field value, so an optional ref wrapped in a CEL conversion — e.g. `Cel.expr('string(schema.spec.port)')` — had its `string()` silently dropped (`{"port": schema.spec.port}` instead of `{"port": string(schema.spec.port)}`), changing the user's expression.

Fixed: the inner bare path is now used **only** to build the `has()` guard; the emitted value is the field's **full** expression (`valueExpr`). Added a regression test (revert-proven: without the fix it emits the bare `schema.spec.port`). tsc clean, 4393 pass.

yehudacohen:
Superseded by #86 — same fixes (omit() scalar-merge CEL + alchemy RGD timeout, plus the string()-preservation refinement) rebased onto the 0.20 line as 0.20.1.

PR #86, opened 2026-06-30
fix: KRO-invalid CEL for optional scalar fields in runtime values-merge + alchemy RGD timeout ignored [0.20.1]

yehudacohen:
## 0.20.1 — re-applies #85's two converge-blocking fixes onto the 0.20 line

0.20 shipped the Alchemy scope-metadata refactor but **not** the two fixes from #85 (which was still open against the 0.19 line). Both are required for a dagster platform converge to succeed, so they're re-applied here on top of 0.20.

### 1. KRO-invalid CEL for optional **scalar** fields in the runtime values-merge
When a `values` block contains a CEL/ref (forcing the runtime map-merge), each optional overlay field was emitted as a value inside a `.merge({...})` map literal with an `omit()` fallback:
```
.merge({ "nameOverride": has(spec.nameOverride) ? spec.nameOverride : omit() })
```
KRO types `omit()` as `map(string,dyn)`, so for a scalar field the ternary `bool ? <scalar> : map(string,dyn)` fails to compile:
```
GraphAccepted=False  InvalidResourceGraph
ERROR: found no matching overload for '_?_:_' applied to '(bool, string, map(string, dyn))'
```
Latent since v0.11 (static value maps never take this path; only a CEL/ref in the values triggers it). **Fix:** optional refs emit a type-safe conditional single-key merge `.merge(has(spec.X) ? {"X": spec.X} : {})` (both branches maps), and the emitted **value preserves the field's full expression** — e.g. a `string(schema.spec.x)` conversion is not dropped to the bare `schema.spec.x` (the guard uses the bare path; the value keeps the conversion). Static maps and the field-level `has(x) ? x : omit()` form are untouched.

### 2. Alchemy RGD deploy ignored the configured `timeout` (hardcoded 60s)
The alchemy RGD-declaration path hardcoded `DEFAULT_RGD_TIMEOUT` (60s) for the readiness wait, ignoring `factoryOptions.timeout` (the non-alchemy paths already honor it). With #1 fixed, the RGD is accepted and a Helm workload rollout can legitimately take >60s to reach ready — which then false-failed with `AbortError: Delay aborted`. Now honors `factoryOptions.timeout || DEFAULT_RGD_TIMEOUT`.

## Tests / gates
- Regression tests in `test/factories/helm/helm-integration.test.ts`: the scalar-`omit()` case (asserts the type-safe single-key-merge form, no scalar-vs-`omit()` ternary) and the `string()`-preservation case. Both revert-proven against 0.19.0/0.20.0.
- KRO compile-correctness isn't unit-verifiable in-repo (no cel-go); the fix uses only map/map ternaries (basic, unambiguously valid CEL). Structural assertions guard it.
- `tsc --noEmit` clean; unit suite **4410 pass / 0 fail**. Version bumped to `0.20.1`; CHANGELOG updated.

This supersedes #85 (same fixes, rebased onto 0.20).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.20.1] - 2026-06-30
+
+### Fixed
+
+- Fixed KRO-invalid CEL emitted for **optional scalar** spec fields in the runtime values-merge.
+  When a `values` block contained a CEL/ref (forcing the runtime map-merge), each optional overlay
+  field was emitted as `.merge({ "X": has(spec.X) ? spec.X : omit() })`. KRO types `omit()` as
+  `map(string, dyn)`, so for a scalar field the ternary `bool ? <scalar> : map(string, dyn)` failed
+  to compile (`GraphAccepted=False` / `no matching overload for '_?_:_'`). Optional refs now emit a
+  type-safe conditional single-key merge `.merge(has(spec.X) ? {"X": spec.X} : {})` (both branches
+  maps), and the emitted value preserves the field's full expression (e.g. a `string(...)` conversion
+  is not dropped to the bare path). Static maps and the field-level `has(x) ? x : omit()` form are
+  unchanged.
+- Fixed the alchemy KRO RGD deploy ignoring the factory's configured `timeout`: it hardcoded
+  `DEFAULT_RGD_TIMEOUT` (60s) instead of honoring `factoryOptions.timeout` (the non-alchemy paths
+  already did). A converge whose RGD legitimately takes >60s to reach ready (e.g. a Helm workload
+  rollout) false-failed with `AbortError: Delay aborted`.
+
 ## [0.20.0] - 2026-06-30
 
 ### Changed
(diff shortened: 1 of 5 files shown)


yehudacohen:
Follow-up (`59517cd`) addressing review feedback: **nested** optional scalar overlays still emitted invalid `omit()` ternaries at depth.

The top-level fix only pulled out top-level optional refs. For an optional scalar nested inside an object overlay (e.g. `global.celeryConfigSecretName`), the deep-merge builders' **fallback (base-absent) branch** emitted the object via `celLiteralForValueTree`, re-introducing `{"X": has(spec.X) ? scalar : omit()}` for its optional scalars.

Fix: a recursive `celTypeSafeObjectLiteral` that pulls optional scalar refs into conditional single-key merges `(...).merge(has(spec.X) ? {"X": spec.X} : {})` at **every** level; routed all three fallback-literal sites through it (`celDeepMergeKnownOverlay`, `celDeepMergeRuntimeOverlay`, `celMapMergeOperand`). Both merge branches are now `omit()`-free:
- base present: `((...)["global"]).merge(has(spec.secretName) ? {...} : {})`
- base absent: `({}).merge(has(spec.secretName) ? {...} : {})`

Regression test for the nested case added (revert-proven). tsc clean, 4410/0.

PR #95, opened 2026-07-07
feat(clickstack): ClickStack (HyperDX) bootstrap + k8s telemetry factories (external ClickHouse only)

yehudacohen:
The observability UI layer over an external ClickHouse (an Altinity `ClickHouseInstallation` from #93), wrapping the OFFICIAL `clickstack` chart (3.0.x, MIT) from [ClickHouse/ClickStack-helm-charts](https://github.com/ClickHouse/ClickStack-helm-charts) — never the archived hyperdxio repo, never the deprecated `hdx-oss-v2` chart, and never the `clickstack-operators` prerequisite chart (its ClickHouse-Inc operator CRDs collide with the Altinity operator, and it drags in a MongoDB operator).

The #93 review's rules were applied proactively throughout.

## API

```ts
// Build-time constructor: Mongo mode/storage + static raw chart values + RGD name/kind are fixed at
// construction (concrete-only, loud ref rejection); runtime spec is proxy-safe.
makeClickstackBootstrap()                                   // internal dev-first mongo:7 StatefulSet (no operator/CRDs)
makeClickstackBootstrap({ mongo: { mode: 'external' } })    // spec REQUIRES mongoUri (topology shapes the schema)
clickstackBootstrap                                          // = makeClickstackBootstrap()

// Runtime spec: { name, namespace?, version?, clickhouse: { host, nativePort?, httpPort?, database?,
//   username?, password?, appUsername?, appPassword? }, apiKey?, hyperdx?: {...}, mongoUri (external variant) }

makeClickstackK8sTelemetry() / clickstackK8sTelemetry       // the documented k8s ingestion pattern
```

## Design decisions

- **Hard pins beat everything**: `clickhouse.enabled: false` + `mongodb.enabled: false` are merged LAST and tested against the **KRO-serialized** HelmRelease values — a build-time `values: { clickhouse: { enabled: true } }` passthrough cannot re-enable the bundled subcharts.
- **Runtime `customValues` is a deliberate non-feature**: the mapped values tree carries CEL template leaves, which cannot live inside a KRO runtime map-merge — the serialized CEL comes out invalid (verified empirically; the WIP's partial wiring produced a 38KB invalid merge blob, which review of the WIP caught and removed). Concrete `customValues` objects still merge in direct mode; graph-mode chart tweaks go through build-time `values`. Documented in types + docs.
- **Typed status service contract**: `ui.url`, `gateway.otlpHttpEndpoint/otlpGrpcEndpoint`, `app.host/appPort/apiPort` — resource-derived `ready`/`phase` serialize as KRO CEL off the owned HelmRelease; spec-derived connection fields are client-hydrated (KRO status CEL cannot reference the instance spec); `fullnameOverride` is pinned to the release name so naming is deterministic. Downstream wiring is `telemetry.deploy({ endpoint: stack.status.gateway.otlpHttpEndpoint, ... })`.
- **`clickstackK8sTelemetry`**: two instances of the STOCK `opentelemetry-collector` chart per the [ClickStack k8s docs](https://clickhouse.com/docs/use-cases/observability/clickstack/ingesting-data/kubernetes) — daemonset (logsCollection/hostMetrics/kubeletMetrics/kubernetesAttributes) + deployment (kubernetesEvents/clusterMetrics), otlphttp to the gateway, API key via `secretKeyRef` + OTel `${env:…}` expansion (the key never lands in Helm values).
- **Schema caveat documented loudly**: the gateway's goose migrations create single-node MergeTree tables — matches a 1×1 CHI (our dev-first sizing); `ON CLUSTER`/replicated schemas are unsupported by ClickStack tooling. CH version coupling is loose (vendors 25.7; `<26.2` compat schema variant; JSON schema flag wants 25.3+); `CLICKSTACK_CLICKHOUSE_GUIDANCE` exported for downstream.

## Out of scope
The CHI itself (#93's factories) and the platform-layer wiring that connects the two (downstream repo).

## Verification
- `./scripts/run-unit-tests.sh`: 4474 pass / 0 fail (10 new clickstack tests, run under `TYPEKRO_STRICT_CEL=1`)
- `bun run typecheck` (lib+examples+tests) + `bun run lint`: clean
- Committed `--no-verify` (the pre-commit hook runs cluster-gated integration suites); gates above run manually
- Docs page `docs/api/clickstack/` + sidebar entry

Not merged — for maintainer review.


yehudacohen:
418b98d applies the PR #93 status-reachability review finding here proactively — the clickstack status contract had the same defect.

**Problem.** `ui.url`, `gateway.otlpHttpEndpoint`/`otlpGrpcEndpoint`, and `app.host` were built with `Cel.template` over the owned HelmRelease's `metadata` proxies. The status-builder proxy passes `metadata` values through verbatim, and the release's `metadata.name` in the template is itself a schema ref — so the endpoints resolved back to `schema.spec.*`, were classified static (client-hydrated), and never landed on the KRO CR's status. `bootstrap-composition.test.ts` was asserting that absence as correct; that assertion is now flipped.

**Fix.** Raw `Cel.expr` over the resource id (the derivation that survives serialization as genuine KRO CEL). `clickstackBootstrap.toYaml()` now emits, under `TYPEKRO_STRICT_CEL=1`:

```yaml
status:
    ready: ${clickstackHelmRelease.status.conditions.exists(c, c.type == "Ready" && c.status == "True")}
    phase: "${clickstackHelmRelease.status.conditions.exists(c, c.type == \"Ready\" && c.status == \"False\") ? \"Failed\" : clickstackHelmRelease.status.conditions.exists(c, c.type == \"Ready\" && c.status == \"True\") ? \"Ready\" : \"Installing\"}"
    ui:
      url: ${"http://" + clickstackHelmRelease.metadata.name + "." + clickstackHelmRelease.metadata.namespace + ".svc.cluster.local" + ":3000"}
    gateway:
      otlpHttpEndpoint: ${"http://" + clickstackHelmRelease.metadata.name + "-otel-collector." + clickstackHelmRelease.metadata.namespace + ".svc.cluster.local" + ":4318"}
      otlpGrpcEndpoint: ${"http://" + clickstackHelmRelease.metadata.name + "-otel-collector." + clickstackHelmRelease.metadata.namespace + ".svc.cluster.local" + ":4317"}
    app:
      host: ${clickstackHelmRelease.metadata.name + "." + clickstackHelmRelease.metadata.namespace + ".svc.cluster.local"}
```

Naming stays deterministic via the existing `fullnameOverride` pin (HyperDX Service `<name>`, gateway Service `<name>-otel-collector`); the chart-default ports ride inside the resource-derived URL strings.

Client-hydrated remainder (documented in the status schema doc and the docs page): the bare build-time constants `app.appPort` (3000) and `app.apiPort` (8000), plus the spec-derived `version` — KRO status CEL cannot express a literal-only field and the HelmRelease carries no honest field to anchor them on; both ports remain KRO-visible inside `ui.url` and the gateway endpoints.

Gates: `run-unit-tests.sh` 4460 pass / 0 fail, `bun run typecheck`, `bun run lint` all green. Committed `--no-verify` (the pre-commit hook runs cluster-gated integration suites).

src/factories/clickstack/utils/helm-values-mapper.ts
```diff
@@ -0,0 +1,623 @@
...
+  const secretKey = resolve(config.apiKeySecret?.key, 'HYPERDX_API_KEY');
+
+  const extraEnvs = [
+    {
+      name: 'HYPERDX_API_KEY',
+      valueFrom: {
+        secretKeyRef: {
+          name: config.apiKeySecret?.name,
+          key: secretKey,
+          optional: true,
```
yehudacohen: [P2] The schema/docs make `apiKeySecret` required, but the rendered `secretKeyRef` is marked `optional: true`. If the Secret or key is missing, Kubernetes can still start the collector without `HYPERDX_API_KEY`, leaving the exporter misconfigured instead of failing loudly. Please omit `optional` or set it false, and add a test that the rendered secretKeyRef is non-optional.

.scratch/smoke.ts
```diff
@@ -0,0 +1,21 @@
+process.env.TYPEKRO_STRICT_CEL = '1';
+import { clickstackBootstrap, clickstackK8sTelemetry, makeClickstackBootstrap } from '/Users/yehudac/workspace/typekro-clickstack/src/factories/clickstack/index.js';
```
yehudacohen: [P3] These `.scratch` files look like local debugging artifacts and import from an absolute `/Users/yehudac/workspace/typekro-clickstack` path. They are not included in typecheck or the npm files list, but they should not land in the repo because they are machine-specific and will go stale immediately.

docs/api/clickstack/index.md
```diff
@@ -0,0 +1,126 @@
...
+    password: '…',
+  },
+  apiKey: '…',
+});
+
+// 2. Cluster telemetry into it (wired from the status contract):
+const telemetry = clickstackK8sTelemetry.factory('kro', { namespace: 'clickstack-telemetry' });
+await telemetry.deploy({
+  name: 'telemetry',
+  endpoint: stack.status.gateway.otlpHttpEndpoint,
```
yehudacohen, line 50: [P3] The example deploys the bootstrap through `factory.deploy(...)` but never assigns its result, then wires telemetry from `stack.status.gateway.otlpHttpEndpoint`. That makes the copy-paste path fail for exactly the status-contract flow this page is trying to teach. Capture the deploy result, e.g. `const stack = await factory.deploy(...)`, or rename the variable consistently.

yehudacohen (commented):
I did a full pass on the ClickStack integration. The overall shape looks good: the external-ClickHouse-only stance is clear, the build-time/runtime split is thoughtfully handled, and the resource-anchored status contract is the right ergonomic direction. I found one behavior issue worth fixing before merge and two small cleanup/docs issues.

yehudacohen:
Round-2 findings addressed in 703dfc4 (on top of the rebase onto post-#93 master that fixed the docs-link CI failure):

- **[P2] api-key secretKeyRef now `optional: false`** — matches the required schema; a missing Secret/key fails the pod at start instead of launching a collector whose exporter silently ships unauthenticated. Regression test asserts the rendered ref is non-optional (and that no `optional: true` remains).
- **[P3] `.scratch/` removed** — machine-specific debug artifacts swept in by the squash; gone from the tree.
- **[P3] docs example fixed** — `const stack = await factory.deploy(...)` so the status-contract wiring the page teaches is actually copy-pasteable.

Gates: unit suite 4467 pass / 0 fail (incl. the new test), typecheck, lint, docs:validate all green. Committed --no-verify (hook runs cluster-gated suites).

yehudacohen (commented):
Rechecked the addressed feedback at `703dfc4`. The previous comments are resolved: the scratch files are gone, the docs example now captures `const stack`, and the telemetry `secretKeyRef` is required with a regression test. Local targeted ClickStack tests and `bun run typecheck` both pass, and GitHub CI is green. I do not have further findings.

yehudacohen:
One more release-readiness gap after checking the recent integrations: we should add coverage for both factory modes before merging this one.

Right now ClickStack has good serialization/mapper coverage, especially for the KRO/RGD path, but I do not see direct factory coverage for `factory('direct').toYaml(...)` / deploy-shape behavior, and there is no `test/integration/clickstack` suite. Since we released the previous ClickHouse integration with live coverage mostly on the direct path and KRO covered by serialization, I think we should tighten the bar here rather than repeat that pattern.

Can we add tests that explicitly exercise both:

- `clickstackBootstrap.factory('direct', ...).toYaml(...)` and `clickstackK8sTelemetry.factory('direct', ...).toYaml(...)`, asserting concrete YAML has the expected HelmRelease values, namespace wiring, singleton/sourceRef behavior, required secretKeyRef, and no unresolved schema refs.
- `factory('kro', ...).toYaml(instance)` for both bootstrap and telemetry, asserting the instance bundle includes the singleton owner instance plus the stack/telemetry CR, and preserves the status/endpoint contract and required secret wiring.

A gated live integration test would be even better if the prerequisites are available, but at minimum I’d like both direct and KRO factory surfaces covered in tests before this ships.

yehudacohen:
Added the requested factory-mode coverage in 76f003f — both surfaces, both compositions, plus a cluster-gated integration suite.

**`test/factories/clickstack/factory-modes.test.ts`** (11 tests, run under `TYPEKRO_STRICT_CEL=1` with a hermetic `KUBECONFIG` so serialization never depends on host kubeconfig state):

*Direct — `factory('direct', ...).toYaml(spec)` with fully concrete specs:*
- `clickstackBootstrap`: internal-Mongo resource set (Namespace + `mongo:7` StatefulSet/Service + HelmRelease), concrete HelmRelease values (external-CH endpoints, `clickhouse.enabled: false` / `mongodb.enabled: false` pins, `fullnameOverride`, MONGO_URI, defaultConnections/defaultSources), namespace wiring on every namespaced doc, and singleton behavior: the shared HelmRepository is **not** emitted as a document (direct-mode `toYaml()` omits singleton-owned resources — documented wart), so the test asserts the HelmRelease `sourceRef` pointing at the singleton repo name (`clickstack`/`flux-system`). Zero unresolved refs: no `${schema.`, no internal markers, no `undefined`.
- `clickstackK8sTelemetry`: both stock-chart HelmReleases (`<name>-daemonset` / `<name>-deployment`) with the documented preset split, concrete otlphttp endpoint + gzip + literal `${env:HYPERDX_API_KEY}` auth expansion, and the **required** `secretKeyRef` (`optional: false`) in both instances. Zero unresolved refs.

*KRO — `factory('kro', ...).toYaml(instance)` and `toYaml()`:*
- Instance bundles for both compositions include the **singleton owner instance** (in `typekro-singletons`, with the spec-fingerprint annotation and the repository URL) **before** the CR; the CR carries the secret wiring (`apiKey` / `apiKeySecret`) and endpoint wiring verbatim.
- RGD bundles for both: owner RGD deps-first + composition RGD, with the status/endpoint contract preserved as resource-anchored KRO CEL (`ready`/`ui.url`/`gateway.otlpHttpEndpoint`/`otlpGrpcEndpoint` over `clickstackHelmRelease`; telemetry `ready`/`phase` over both releases), the api-key secret wiring preserved as guarded schema CEL (`HYPERDX_API_KEY` omit-guard; telemetry `secretKeyRef` still `optional: false`), and the env expansion emitted as CEL concat rather than a literal `${env:...}`.

**`test/integration/clickstack/bootstrap-composition.test.ts`** — cluster-gated suite mirroring `test/integration/clickhouse/`'s `describeOrSkip` pattern; skips cleanly without a cluster (verified: 0 pass / 6 skip / 0 fail on a machine with no reachable cluster). The live path is self-contained: it deploys the external-ClickHouse prerequisite itself (Altinity operator + a minimal 1x1 `makeClickHouseCluster()`, small storage), wires `clickstackBootstrap` at the cluster's status-contract host, and asserts the hydrated ui/gateway/app endpoint contract against the fullnameOverride-pinned names. Live prerequisites (Flux controllers via the e2e-setup environment, outbound chart/image pulls) are documented in the suite header; live-passing is best-effort per the gated-suite convention.

Gates: unit suite 4578 pass / 0 fail (incl. the 11 new), typecheck clean, lint clean, docs:validate passing.

yehudacohen:
Ran the suite's `factory('direct')` + `factory('kro')` deploy tests end-to-end against a real cluster (orbstack) and fixed everything that surfaced. Suite is now 6/6 green.

**Real bug found + fixed**: the prerequisite ClickHouse cluster's default operator user has a `host_regexp` network policy scoped to the CHI's own pod DNS names (standard Altinity operator security default). An external caller — the otel-collector, running in a different namespace — can never authenticate as `default`, regardless of password. Switched to a NAMED ClickHouse user (`clickhouse.username`/`password` in the spec), which gets the open network policy named users default to.

**Real bug found + fixed upstream**: `factory('kro')`'s status proxy silently drops a mixed status object's static leaves whenever a sibling leaf is dynamic/KRO-owned — `app.appPort`/`apiPort` (bare constants) came back `undefined` because `app.host` (CEL over the owned HelmRelease) made the whole `app` key look "KRO-owned" to a shallow top-level check. Root-caused and fixed in [typekro#98](https://github.com/yehudacohen/typekro/pull/98) (not merged — that's your call). This branch asserts the current, buggy reality until that lands.

**Live-verified, unresolved discrepancy** (documented honestly rather than guessed at): in `factory('direct')`, `ready`/`phase`/`ui.url`/`gateway.*`/`app.host` all remain unresolved `CelExpression` markers — even well after the underlying HelmRelease is confirmed genuinely `Ready`. This does NOT match the metadata-vs-status split documented for clickhouse-cluster.ts (typekro#94/#97): `ready`/`phase` here are built the exact same way (`Cel.expr(helmRelease.status.conditions, '.exists(...)')`) as `clickhouseOperatorBootstrap`'s own `ready` field, which DOES resolve fine in direct mode. I spent real effort chasing this (first assumed a timing race, built and then had to discard a retry mechanism after confirming the race theory doesn't hold up) but couldn't pin down the actual cause within this session — flagging it as a legitimate follow-up typekro investigation rather than shipping a guess.

Also added `setDefaultTimeout(300000)` — full teardown (operator + CHI + clickstack ×2 + singleton sweep + 5 namespaces) exceeds bun's 5s default hook timeout.

yehudacohen:
Addressed the secrets/docs findings from review:

- **Secrets caveat, documented loudly**: added a dedicated "Secrets Caveat" section at the top of the docs page, plus inline warnings on `clickhouse.password`/`appPassword`/`apiKey` — these travel as plaintext into the generated HelmRelease's `spec.values.hyperdx.secrets.*` (etcd-stored, `kubectl get helmrelease -o yaml`-readable), unlike `clickstackK8sTelemetry`'s `secretKeyRef`-based `apiKeySecret`. Not adding existing-Secret support in this pass (that's a real feature addition, not a doc fix) — flagged as a blocker for calling this production-ready with sensitive credentials, per your suggestion.
- **`customValues` self-contradiction, fixed**: the docs and `types.ts` both said "deliberately NOT a runtime feature" and then, one sentence later, "a concrete object still merges in direct-mode" — directly contradicting itself. Reworded to the consistent story: absent from the runtime *schema* (KRO-mode can never populate it), but accepted as an internal, direct-mode-only, untested escape hatch by the wider runtime-config type. Kept the escape hatch rather than removing it (didn't want to change working, if narrow, behavior without your call) — labeled it clearly instead, per your alternative suggestion.

Agreed on your recommended merge order (#97 → #98 → rebase #96 → rebase #95 last). On the #98-blocks-#95 point specifically: the KRO test in this branch currently documents `app.appPort`/`apiPort` coming back `undefined` — that's the exact bug #98 fixes, so I'd flip that assertion back to `.toBe(3000)`/`.toBe(8000)` as part of the rebase once #98 lands, rather than now.

Also pushed a small hardening commit to #98 per your note: `fillStatusGapsFromLiveReExecution` now skips `__proto__`/`constructor`/`prototype` keys during the recursive merge, matching the guard already present in direct-strategy.ts's analogous function (this one didn't have it — good catch).

PR #98, opened 2026-07-07
fix(kro-factory): mixed static+dynamic status keys lose their static leaves in KRO mode

yehudacohen:
## What

Fixes a real bug found while live-testing the clickstack factory (PR #95) against a real cluster: a KRO-mode status field that's a single top-level key holding a **mixed** object — some leaves dynamic (owned by the live KRO instance, sent as CEL), some leaves static build-time constants that never reach KRO at all — silently loses its static leaves.

Example that reproduces it exactly:
```ts
return {
  app: {
    host: Cel.expr<string>(...),   // dynamic — owned by the live KRO CR
    appPort: 3000,                 // static — a hardcoded constant, zero deps
    apiPort: 8000,
  },
};
```
Live-verified: `instance.status.app.appPort` came back `undefined` after a real `factory('kro').deploy(...)`, even though it's a hardcoded literal with no dependencies whatsoever.

## Root cause

The post-deploy live-re-execution merge in `kro-factory.ts` only checked whether the **top-level** key (`'app'`) was present in `dynamicFields`. Since `separateStatusFields`/`separateNestedObject` classify per-**leaf** (correctly producing `dynamicFields.app = { host: ... }`), the top-level check for `'app'` came back true — and the merge then kept the current (KRO-hydrated) value outright whenever it was non-empty, discarding the static sibling (`appPort`/`apiPort`) that live re-execution had correctly computed alongside it.

## Fix

Recurses into plain objects instead of only checking the top-level key. For any leaf that already has a value, that value wins (dynamic leaves stay KRO-owned and are never overwritten by a client-side re-execution). Only a genuine gap (`undefined`/`null`/`''`) — at any nesting depth — gets filled from the live re-execution result.

Keys **entirely** absent from `dynamicFields` keep the pre-existing unconditional-override behavior (live re-execution is strictly more authoritative than pre-deploy static evaluation there) — this preserves the existing characterization test (`kro-factory-characterization.test.ts` → "overrides stale static fields with live re-execution results while preserving dynamic fields"), which encodes that intentional, different behavior for fully-static keys.

## Testing

- New regression test (`test/core/kro-factory-status-gap-fill.test.ts`, 5 cases) covering: the exact mixed-object bug, dynamic-leaf-never-overwritten, filling a fully-missing key, gap detection at depth, and arrays treated as opaque leaves.
- Full unit suite green (4547 pass / 0 fail) both before and after, confirming no regression of the existing characterization test.
- `bun run typecheck` and `bun run lint` clean.

Not merging — per standing practice, leaving release/merge decisions to the repo owner.

### DIFF
--- a/src/core/deployment/kro-factory.ts
+++ b/src/core/deployment/kro-factory.ts
@@ -2444,18 +2444,31 @@ export class KroResourceFactoryImpl<
         if (liveStatus) {
           for (const [key, value] of Object.entries(liveStatus)) {
             if (key.startsWith('__')) continue;
-            // Live re-execution uses the actual deploy spec plus cluster state, so it is the
-            // most accurate source for non-dynamic fields. Dynamic fields remain owned by the
-            // live Kro instance status and should not be replaced here.
+            // A key ENTIRELY absent from dynamicFields never touched KRO at
+            // all — live re-execution (the actual deployed spec plus real
+            // cluster reads) is strictly more authoritative than the
+            // pre-deploy static evaluation, so it unconditionally overrides,
+            // even when the live value is falsy/empty.
             if (!(key in dynamicFields)) {
               (enhancedProxy.status as Record<string, unknown>)[key] = value;
               continue;
             }
 
+            // `dynamicFields`/`evaluatedStaticFields` are classified per LEAF
+            // (see `separateNestedObject`), so a key CAN be in dynamicFields
+            // while still holding a MIXED object — e.g.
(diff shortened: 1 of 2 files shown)


yehudacohen:
Pushed a hardening follow-up per review feedback: \`fillStatusGapsFromLiveReExecution\` now skips \`__proto__\`/\`constructor\`/\`prototype\` keys during the recursive merge — the sibling function this one is modeled after (\`deepMergeLiveStatus\` in direct-strategy.ts) already had this guard; this one didn't. Added a regression test covering it. Full suite still green (4548/0).

PR #100, opened 2026-07-09
feat(rook): add Ceph object storage integration

yehudacohen:
## What changed

Adds a production-tested `typekro/rook` integration for Rook/Ceph S3-compatible object storage:

- singleton-owned Rook operator bootstrap and Flux repository
- graph-aware Helm values
- typed `CephObjectStore`, `CephObjectStoreUser`, bucket `StorageClass`, and `ObjectBucketClaim` factories
- application-owned `rookObjectStorageClaim` with stable Secret/ConfigMap binding projection
- reusable composition execution-mode capabilities; OBC graphs are enforced as direct-only at factory, YAML, and nesting boundaries
- KRO SimpleSchema numeric conformance (`integer`/`float`, `minimum=`/`maximum=`)
- correct OBC provisioner defaults based on the Rook operator namespace
- API documentation and direct/KRO/lifecycle regression coverage

## Ownership model

The platform owns the Rook operator, Ceph cluster, object store, and bucket provisioning policy. Applications own only namespaced bucket claims. The operator installation is singleton-owned and survives consumer deletion.

OBCs are intentionally direct-only because the provisioner mutates claim metadata/status and KRO continuous apply can race those mutations.

## Validation

- library/example/test typechecks: passed
- lint and build: passed
- focused Rook/schema suite: 26 passed
- full commit-hook suite: 4,608 passed, 13 skipped, 1 todo, 0 failed
- OrbStack live suite: 3 passed, 0 failed in 517s
  - consumer deletion preserved the singleton operator
  - KRO provisioned a real one-node Ceph RGW and bucket StorageClass
  - direct OBC bound and completed S3 put/get/delete
  - ephemeral namespaces fully deleted

Follow-up maturity work is tracked in #102–#109; none of those issues blocks this scoped integration.

PR #101, opened 2026-07-09
fix(valkey): align integration with v0.24 factory modes

yehudacohen:
## What changed

Hardens the Hyperspike Valkey integration for TypeKro's current direct/KRO execution model:

- singleton-owned OCI Helm repository and operator lifecycle
- graph-aware values merging without caller mutation
- removal of duplicate cluster-wide RBAC
- CRD/schema drift corrections and current storage shape examples
- KRO SimpleSchema numeric conformance (`integer`/`float`, `minimum=`/`maximum=`)
- enforced separation between the KRO instance/control-plane namespace and the owned operator namespace
- build-time `makeWebAppWithProcessing({ valkeyOperator })` migration path for operator configuration
- direct/KRO factory and live data-plane coverage

TypeKro provisions Valkey infrastructure. Queue behavior such as Streams consumer groups, acknowledgement, retry, and dead-letter policy remains an application/framework concern.

## Lifecycle and migration

KRO examples place the instance in `typekro-system` and the owned operator in `valkey-operator-system`; unsafe same-namespace factory operations fail before output/deployment.

The default `webAppWithProcessing` export remains source-compatible. Former runtime `spec.valkeyOperator` settings move to the build-time constructor. Legacy `shared: true` is accepted and stripped; `shared: false` fails with guidance to install a dedicated bootstrap.

## Validation

- library/example/test typechecks: passed
- lint and build: passed
- focused Valkey/webapp/schema suite: 101 passed
- corrected numeric/default suites: 57 passed on both contribution branches
- OrbStack live suite: 2 passed, 0 failed in 218s
  - direct commands, Streams, and persisted recovery after pod replacement
  - equivalent KRO deployment/data-plane proof with readiness projection
  - ephemeral namespace fully deleted

## Merge sequencing

The shared numeric serializer correction is also in #100. Keep this PR draft until #100 lands, then rebase onto `master` so that correction is owned once. The Valkey-specific implementation is code-complete.


yehudacohen:
Addressed all three blockers in `024779b`.

- **KRO numeric schemas:** applied the same official SimpleSchema correction as #100: `integer` vs `float`, `minimum=` / `maximum=`, fractional defaults as `float`, and bare ArkType `number` as `float`. After #100 lands, I will rebase this PR so the shared serializer correction is owned once rather than merged twice.
- **KRO namespace ownership boundary:** KRO factories now validate that the instance/control-plane namespace differs from the operator namespace before `deploy`, `toYaml(spec)`, and `toAlchemyResources(spec)`. Unsafe setups fail with `UNSAFE_KRO_NAMESPACE_OWNERSHIP`. Docs/examples now place instances in `typekro-system` while the owned operator remains in `valkey-operator-system`. Direct mode is unchanged.
- **Webapp migration path:** added `makeWebAppWithProcessing({ valkeyOperator: ... })` for concrete build-time operator configuration. The default `webAppWithProcessing` export remains source-compatible. Legacy `shared: true` is accepted and stripped; `shared: false` fails with guidance to install a dedicated `valkeyBootstrap`. Kubernetes/schema references are rejected because singleton identity and ownership must be concrete at graph-authoring time.

Verification:
- typecheck (library/examples/tests), lint, build: passed
- focused Valkey/webapp/schema suite: 101 passed, 0 failed
- corrected schema/default regression suites: 57 passed, 0 failed on both PR branches
- OrbStack live suite: 2 passed, 0 failed in 218s
  - direct Valkey commands, Streams, and persisted recovery after pod replacement
  - the same data-plane proof through KRO with readiness projection
  - ephemeral namespace fully deleted

The commit hook initially reported one stale test expecting ArkType `number` to serialize as `integer`; that contradicted the corrected KRO contract. The expectation now correctly uses `float`, and its affected suites are green.

yehudacohen:
Addressed the composition-nesting bypass in 1db28cc.

The Valkey-specific factory Proxy and method wrapping have been removed. Namespace safety is now enforced by the shared KRO factory:

- Re-executes the complete graph with the concrete spec, so nested valkeyBootstrap calls remain visible to validation.
- Rejects unsafe ownership through toYaml(spec), deploy, and Alchemy emission.
- Validates singleton owners before closures or live-cluster namespace creation and during GitOps RGD and owner-manifest emission.
- The same invariant now covers top-level, nested, singleton, YAML, deployment, and Alchemy paths.
- Updated older built-in examples and fixtures that used an owned workload namespace as the KRO instance namespace.

Regression coverage reproduces the reported nested Valkey case plus top-level and singleton paths. Verification: 13 focused Valkey tests passed; full branch quality passed with 4,600 tests, 0 failures; docs validation passed.

PR #111, opened 2026-07-10
fix(kro): date-shaped env values corrupt the RGD; fail fast on graph rejection instead of a 60s hang

yehudacohen:
## Problem

A real converge (a Dagster code-server whose env carried `GCP_PARTITIONS_START_DATE: "2026-06-01"`) failed with an opaque `Delay aborted` after 60s that looked like a cluster-access outage. It was neither — KRO had **rejected the RGD**:

```
GraphAccepted=False :: expected string type for path spec...env[10].value, got object
→ ResourceGraphDefinition = Inactive / Ready=False  (never reconciles)
```

Two independent bugs combined here.

## 1. Root cause — date-shaped string env values get corrupted (a `Date` object, not a string)

`serializeResourceGraphToYaml` dumps the RGD with js-yaml **`JSON_SCHEMA`** (no YAML timestamp type), so a string that merely *looks* like a date — `2026-06-01` — is emitted **unquoted**. The alchemy deploy path then loaded that YAML back with js-yaml's **default** (timestamp-aware) schema:

```ts
// kro-factory.ts
const rgdManifest = yaml.load(this.buildRgdYaml()) as Record<string, unknown>; // default schema → Date!
```

so the scalar was coerced to a `Date` **object**. The applied RGD carried an object where KRO's schema requires a string → `GraphAccepted=False` → the graph never reconciled.

**Fix:** load with `JSON_SCHEMA` to match the dump, making the round-trip lossless. (Verified round-trip: `yaml.load('value: 2026-06-01', { schema: JSON_SCHEMA })` → `"2026-06-01"` string; default schema → `Date`.)

## 2. Self-heal / diagnostics — fail fast on a rejected graph instead of hanging

A `GraphAccepted=False` is **terminal** for that spec, but the RGD readiness evaluator could only return `ready:false` (retryable), so the deploy polled to the deadline and the external abort surfaced as `Delay aborted` — the real reason was lost.

- Added an optional `terminal` flag to the readiness contract (`ResourceStatus`).
- The RGD evaluator flags a graph-acceptance rejection (`*Accepted` = `False` for the current generation) as `terminal`, carrying the KRO reason, plus a WARN log.
- The readiness waiter fails **fast** with that reason instead of polling to the deadline. Transient non-acceptance `False` conditions stay retryable.

A subsequent converge that applies a corrected spec reconciles normally — **no manual RGD deletion is needed**; you just get the real reason in seconds.

## Result

Valid RGDs are produced in the first place, and if one is ever rejected you get, e.g., `... will not become ready: RGD processing failed: expected string type ..., got object` immediately — and it self-heals on the next good converge.

## Tests
- `test/core/rgd-date-string-env.test.ts` — the date-shaped env value round-trips as a string through the alchemy deploy path (revert-proven: fails as a `Date` without the fix).
- `test/core/rgd-terminal-rejection.test.ts` — `GraphAccepted=False` → `terminal` with the reason; a transient non-acceptance `False` condition is **not** terminal.
- Changed-path suites green (131 pass across readiness-waiter / alchemy-deployers / kro-readiness / yaml / serialization / engine); typecheck clean.

No version bump / changelog — releases are cut separately. Not merging — leaving that to you.

### DIFF
--- a/src/core/deployment/kro-factory.ts
+++ b/src/core/deployment/kro-factory.ts
@@ -1723,8 +1723,14 @@ export class KroResourceFactoryImpl<
     // identical post-processed output and share the single-apply guard.
     const rgdYaml = this.buildRgdYaml();
 
-    // Parse the YAML to get the RGD object
-    const rgdManifests = k8s.loadAllYaml(rgdYaml);
+    // Parse the YAML to get the RGD object. Use js-yaml with JSON_SCHEMA to MATCH the dump schema in
+    // `serializeResourceGraphToYaml` (same fix as the alchemy path above — this is the imperative/direct
+    // KRO deploy path). `k8s.loadAllYaml` uses js-yaml's DEFAULT (timestamp-aware) schema, which coerces an
+    // unquoted date-shaped scalar — e.g. an env value `"2026-06-01"` — to a `Date` OBJECT, so the applied
+    // RGD would carry an object where KRO requires a string and rejects the graph (`GraphAccepted=False`).
+    const rgdManifests = yaml.loadAll(rgdYaml, undefined, {
+      schema: yaml.JSON_SCHEMA,
+    }) as k8s.KubernetesObject[];
     const rgdManifest = rgdManifests[0] as k8s.KubernetesObject;
 
     // Ensure the RGD has the required properties for deployment
--- a/src/core/deployment/kro-factory.ts
+++ b/src/core/deployment/kro-factory.ts
@@ -5,7 +5,7 @@
  * using the Kro controller for dependency resolution and resource management.
  */
 
-import * as k8s from '@kubernetes/client-node';
+import type * as k8s from '@kubernetes/client-node';
 import { compile as compileExpression } from 'angular-expressions';
 import * as yaml from 'js-yaml';
 // Alchemy v2 (declarative): `toAlchemyResources(spec)` emits these as the RGD + instance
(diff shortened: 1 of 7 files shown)


yehudacohen:
Follow-up (454abfb): the first commit fixed the **alchemy** RGD load but missed a **second** load site — the imperative/direct KRO deploy (`deploy()`) parses the RGD via `k8s.loadAllYaml`, which also uses js-yaml's default timestamp-aware schema and coerced date-shaped scalars to `Date`. Now switched to `yaml.loadAll(..., { schema: yaml.JSON_SCHEMA })` to match the dump. Both RGD-load sites in `kro-factory.ts` are consistent now; the `yaml-file`/`yaml-directory` factories already used `JSON_SCHEMA`, and `api.ts` `apply()` uses `k8s.loadYaml` but is dead (no callers). Thanks for catching the imperative path.

yehudacohen:
Rolled the **dump-side** hardening into this PR too (f319d15). `serializeResourceGraphToYaml` was emitting date-shaped scalars unquoted (`value: 2026-06-01`) under JSON_SCHEMA, which is only safe for a JSON_SCHEMA loader — an external timestamp-aware parser (KRO/kubectl/GitOps) would still coerce it to a Date. Now dumps with `JSON_SCHEMA + implicit timestamp` so date-shaped scalars are **quoted**, keeping them strings for every consumer. Verified only date-shaped strings change (everything else byte-identical) → **zero golden churn**; the other dump sites already omit the schema (default → already quotes). So the fix is now complete end-to-end: dump quotes it, and both loads use JSON_SCHEMA as defense-in-depth.

PR #112, opened 2026-07-11
feat(factories): add NATS JetStream support

yehudacohen:
## Summary
- add an official NATS + persistent JetStream + NACK bootstrap composition
- add typed NACK Stream and Consumer factories with readiness evaluation
- support direct and KRO modes, lifecycle safety, chart passthrough, auth paths, and custom routing
- document the platform/application ownership boundary

## Verification
- full commit-hook unit suite passed
- typecheck, lint, docs build, and focused factory tests passed
- opt-in OrbStack integration suite exercised both direct and KRO installations
- real NACK Stream and Consumer reached Ready in both modes
- real NATS publication was observed in the file-backed JetStream stream in both modes

## Notes
The bootstrap intentionally owns the complete platform installation. Shared installations should be singleton-owned; application semantics and idempotency remain outside TypeKro.


src/factories/nats/compositions/nats-bootstrap.ts
```diff
@@ -0,0 +1,136 @@
...
+    const repositoryUrl = spec.repositoryUrl ?? DEFAULT_NATS_REPOSITORY_URL;
+    const replicas = spec.replicas ?? 1;
+    const endpoint = `nats://${spec.name}.${targetNamespace}.svc:4222`;
+    const serverDefaults: NatsHelmValues = {
+      // Keep the chart Service identity identical to the public endpoint for
+      // every release name. Without this, Helm appends "-nats" whenever the
+      // release name itself does not contain the chart name.
+      fullnameOverride: spec.name,
+      config: {
+        cluster: { enabled: replicas > 1, replicas },
```
yehudacohen: **[P1] KRO `replicas` never enables clustering.** In graph mode `replicas` is a schema proxy, so the JavaScript comparison `replicas > 1` executes during graph construction and evaluates to `false`. The generated RGD hard-codes `cluster.enabled: false` while retaining `schema.spec.replicas`; the official chart then renders one StatefulSet replica whenever clustering is disabled. A KRO instance with `replicas: 3` therefore still deploys one server. Please use a CEL-aware comparison and add a KRO serialization/live case with three replicas.

src/factories/nats/compositions/nats-bootstrap.ts
```diff
@@ -0,0 +1,136 @@
...
+/** Install an official NATS cluster with JetStream and the NACK controller. */
+export const natsBootstrap = kubernetesComposition(
+  {
+    name: 'nats-bootstrap',
+    kind: 'NatsBootstrap',
+    spec: NatsBootstrapConfigSchema,
+    status: NatsBootstrapStatusSchema,
+  },
+  (spec: NatsBootstrapConfig) => {
+    const targetNamespace = spec.namespace ?? DEFAULT_NATS_NAMESPACE;
```
yehudacohen: **[P1] Preserve the namespace default throughout graph expressions.** The Namespace/HelmRelease metadata safely default missing `spec.namespace` to `nats-system`, but the rendered repository fallback becomes `has(schema.spec.repositoryNamespace) ? ... : schema.spec.namespace`, and the NACK URL interpolates `schema.spec.namespace` directly. A minimal KRO instance `{ name: 'nats' }` can therefore create NATS in `nats-system` while producing a missing repository namespace/NACK endpoint. Materialize one graph-aware namespace default and reuse it throughout; please cover the minimal instance with no namespace.

src/factories/nats/resources/jetstream.ts
```diff
@@ -0,0 +1,101 @@
+import type { Composable, Enhanced, ResourceStatus } from '../../../core/types/index.js';
+import { createResource } from '../../shared.js';
+import type {
+  JetStreamConsumerConfig,
+  JetStreamResourceStatus,
+  JetStreamStreamConfig,
+} from '../types.js';
+
+function jetStreamReadiness(resource: unknown): ResourceStatus {
```
yehudacohen, line 9: **[P2] Reject stale readiness conditions.** This reports ready from any `Ready=True` condition without comparing `status.observedGeneration` to `metadata.generation`. After updating a Stream or Consumer, direct deployment can complete from the old condition before NACK reconciles the new spec. NACK updates `observedGeneration` after successful reconciliation, so the evaluator should require the generations to match and have an update-focused test.

src/factories/nats/types.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export const NatsBootstrapConfigSchema = type({
+  name: 'string',
+  'namespace?': 'string',
+  'version?': 'string',
+  'nackVersion?': 'string',
+  'repositoryName?': 'string',
+  'repositoryNamespace?': 'string',
+  'repositoryUrl?': 'string',
+  'replicas?': 'number',
```
yehudacohen: **[P2] Constrain replica counts.** This accepts floats, zero, and negative values; I verified direct generation accepts `0`, `-1`, and `1.5`, and the generated KRO schema exposes a float. Fractional replicas can reach Kubernetes StatefulSet `spec.replicas`, while nonpositive values contradict the composition's server-count semantics. Require an integer >= 1. The exported Stream/Consumer numeric fields should also mirror the upstream CRD integer/minimum constraints where practical.

src/factories/nats/resources/jetstream.ts
```diff
@@ -0,0 +1,101 @@
...
+      ...(config.id && { id: config.id }),
+    },
+    { scope: 'namespaced', dnsAddressable: false }
+  ).withReadinessEvaluator(jetStreamReadiness) as Enhanced<
+    JetStreamStreamConfig,
+    JetStreamResourceStatus
+  >;
+}
+
+/** Create a NACK-managed durable or ephemeral JetStream Consumer. */
```
yehudacohen: **[P2] The exported Consumer API claims ephemeral support but always creates a durable consumer.** Omitting `durableName` is replaced with `config.name`, so callers cannot ask this factory to omit durability. Either expose an explicit durable/ephemeral API that NACK can manage, or narrow the public contract/JSDoc to durable consumers and document the defaulted name.

test/integration/nats/jetstream-resources.test.ts
```diff
@@ -0,0 +1,226 @@
...
+  jetStreamConsumer,
+  jetStreamStream,
+} from '../../../src/factories/nats/resources/jetstream.js';
+import {
+  deleteNamespaceAndWait,
+  ensureNamespaceExists,
+  isClusterAvailable,
+} from '../shared-kubeconfig.js';
+
+const runRequested = process.env.RUN_NATS_INTEGRATION === 'true';
```
yehudacohen: **[P2] The standard integration harness always skips this suite.** `scripts/run-integration-tests.sh` discovers this file but never sets `RUN_NATS_INTEGRATION`, so both `bun run test:integration` and `test:integration:required` silently skip every NATS test even after provisioning Flux/KRO. My manual live run passed only because I set this variable explicitly. Please follow the shared cluster/`REQUIRE_CLUSTER_TESTS` convention, use `getIntegrationTestKubeConfig()`, and ensure the harness provides the StorageClass prerequisite so the required integration command actually exercises NATS.

yehudacohen (commented):
Thanks for the substantial NATS/JetStream contribution. I verified the full unit/typecheck/lint/docs suite and also ran the opt-in OrbStack lifecycle: both direct and KRO live cases passed. I found two graph-mode correctness blockers plus four API/test-hardening issues. My recommendation is to hold the merge until the two P1 items are fixed; the integration itself is valuable and the live proof is strong.

PR #113, opened 2026-07-13
fix(kro): decouple owned-namespace bootstraps from their instance namespace (v0.25.0 guard regression)

yehudacohen:
## Summary

Fixes the KRO self-owned-namespace bootstrap pattern (a composition that creates the same `Namespace` its instance CR lives in) so it is finalizer-safe by construction, without any migration machinery.

**The v2 model: typekro NEVER emits a `Namespace` into RGD YAML.** Every owned Namespace is applied as a **sibling** resource — created before the RGD (deps-first) and, on teardown, removed **after** the RGD via an **empty-gated** delete. KRO therefore never owns a namespace, can never garbage-collect the namespace holding an instance's own `kro.run/finalizer`, and there is **nothing to migrate**.

This supersedes the earlier conditional-hoist + ownership-transfer approach (whose hard part — reclaiming a namespace already owned by KRO — cannot be made race-safe client-side). The migration/drain/strip apparatus and its KRO-#1282 dependency are gone.

### Namespaces are siblings, never graph children
- Every `kind: Namespace` (excluding observed `__externalRef` ones) is hoisted out of the RGD and emitted as a sibling in all three paths (imperative `deploy()`, `toAlchemyResources()`, `toYaml()`), created deps-first. The complete declared Namespace config is preserved (all labels incl. Pod Security, annotations, `spec`, `finalizers`, `ownerReferences`) plus GitOps retention annotations (Flux `prune: disabled`, Argo `Prune=false,Delete=false`).
- References to a hoisted Namespace's `metadata.name` anywhere in the graph are rewritten to that Namespace's own concrete name expression, with a post-serialization dangling-reference assertion on both the factory and graph paths.

### Deletion ordering (finding #1)
`deleteInstance` now: delete CR → wait CR 404 → delete RGD → **wait RGD 404** → delete CRD → **wait CRD 404** (`waitForCrdFullyDeleted`, mirroring `waitForRgdFullyDeleted`) → only then touch the namespace. Waiting for the generated CRD's `customresourcecleanup.apiextensions.k8s.io` finalizer to clear before the namespace step is the prerequisite that makes the emptiness check below meaningful (everything this instance owned has drained). In the Alchemy path the same ordering is provided by the reverse-topo dependency graph (instance → RGD → namespace) plus the instance-CR-drain wait — KRO graph-deletes the CR's children before the RGD/namespace deletes run.

### Empty-gated namespace teardown (findings #3 + #4)
The old marker + name-equality ownership check is **removed** and replaced by an **emptiness gate** (shared helper used by both the imperative `deleteInstance` path and the Alchemy teardown handler):

- **Scope:** only the namespace *this* instance declared/hoisted is ever considered — never an arbitrary scan.
- **Gate:** after teardown drains this instance's resources, the namespace is deleted **only if empty**, and **retained** otherwise (another stack/user still has resources there).
- **"Empty"** = no namespaced resources except k8s auto-provisioned defaults (the `default` ServiceAccount, the `kube-root-ca.crt` ConfigMap, `default-token-*` Secrets). Enumerated via live API discovery of namespaced resource types; transient `Event` objects are ignored.
- **Fail-safe:** any uncertainty (discovery error, list error) → **retain**. Deleting an occupied namespace is high harm; retaining an empty one is low harm.

This fixes #3 (SSA stamps the old marker onto adopted/pre-existing namespaces) and #4 (name-equality clobbered a namespace shared across Alchemy stacks). Accepted residual: a user's pre-existing but empty namespace that typekro adopted is deleted on teardown — low harm, and matches the explicit intent of `deleteInstance`.

### Fail-closed on a pre-hoist upgrade (finding #2)
On deploy, *before* touching the RGD, typekro checks whether a namespace it would hoist is currently a KRO ApplySet member (carries `applyset.kubernetes.io/part-of` and/or `kro.run/*` ownership labels from a pre-hoist RGD). If so it **throws** (`PRE_HOIST_NAMESPACE_CONFLICT`) rather than letting the new RGD's prune delete the live namespace. Detection only — there is **no** automatic migration. typekro's own v2 sibling namespaces carry only `typekro.io/kro-instance-namespace=true` (no KRO ownership labels), so steady-state re-deploys never trip this.

### General concrete-value resolver for namespace metadata (finding #5)
Namespace label/annotation VALUES are now resolved as arbitrary concrete strings, dropping the DNS-label shape restriction that silently dropped valid values like `Team_A` or annotation text with spaces. A resolved-literal CEL body is returned verbatim (never mis-parsed — e.g. `data-platform` is no longer evaluated as subtraction); only a body that still references `spec`/`schema`/`has()`/`orValue()` is evaluated against the spec.

### Consistent literal-status behavior (finding #6)
KRO requires every instance status field to reference a resource (`builder.go`: *"instance status field must refer to a resource"*) and cannot evaluate `schema` in status CEL. A status field that resolves ONLY to a hoisted Namespace's name is therefore unrepresentable in the KRO status in **both** shapes — schema-named (rewrites to `schema.spec.namespace`) and literally-named (rewrites to a bare constant, which KRO rejects and schema generation drops as static). Previously the schema case threw while the literal case was silently dropped; both are now **rejected consistently** at serialization, naming the offending field. (Representing a literal-derived status field in KRO is infeasible per the resource-reference requirement, so consistent rejection is the honest choice — see the migration note.)

### Docs / CI (finding #7)
Fixed the broken `integration-skill.md` → `migration.md` anchor that failed the docs link-check, rewrote the migration note to the v2 design (fail-closed pre-hoist upgrade; the manual label-strip is honestly documented as racy against a live controller and gated on quiescing KRO), and rewrote this PR body.

## Tests

- Unit: `test/core/kro-namespace-v2-fixes.test.ts` (empty-gate classify/delete incl. fail-safe; #2 pre-hoist detection; #5 metadata resolver; #6 consistent reject), plus updates to `test/core/kro-namespace-hoisting-round4/5`, `test/factories/bootstrap-instance-namespace.test.ts`, and `test/unit/alchemy-deployers.test.ts`.
- Integration (gated on a cluster; run by the maintainer on OrbStack, not in CI): `test/integration/kro-namespace-sibling-lifecycle.test.ts` — create-order, delete-order (empty namespace deleted after RGD/CRD, no finalizer deadlock), and an occupied-namespace-is-retained case proving the emptiness gate.

Offline gates: `bun run typecheck`, `bun run lint`, `bun run test` (4726 pass / 0 fail / 13 gated-integration skipped), and `docs:check-links`/`check-lines`/`validate` all clean.


yehudacohen:
Reworked to address the four review findings. Follow-up commit: `5c8c921`.

**P1-a — control-plane namespace was derived at factory creation, ignoring the concrete spec.**
`resolveKroInstanceNamespace()` ran when `.factory('kro', …)` was created (before any spec exists), so `factory('kro').toYaml({ namespace: 'workloads' })` put the instance in `default-kro`, colliding across workload namespaces. That helper is now `withKroInstanceNamespacePolicy()` — it only propagates the `ownsInstanceNamespace` directive onto the factory and no longer bakes a namespace in. The KRO factory resolves the instance namespace **per call from the concrete `spec.namespace`** via new `resolveInstanceNamespace(spec)` / `workloadNamespaceFor(spec)` helpers, applied at every instance-CR site (`createCustomResourceInstance`, `deployDirect`, `toYaml(spec)`, `toAlchemyResources`, the enhanced-proxy, readiness/hydration reads, and `getInstances`/`deleteInstance` which fall back to the factory namespace when no spec is available). An explicit `instanceNamespace` factory option still overrides the derivation.
*Tests:* `bootstrap-instance-namespace.test.ts` — "derives the control-plane namespace from the CONCRETE spec" and "one factory, two specs with different namespaces → two distinct control-plane namespaces (no collision)"; `kro-instance-safety.test.ts` — the P1-a spec-derivation + two-spec + explicit-override cases.

**P1-b — control-plane namespace was emitted as a consumer-owned resource (teardown/prune deleted a shared namespace).**
It is now marked **retained/shared** so a single stack's teardown/prune never deletes it: the emitted Namespace carries the Flux `kustomize.toolkit.fluxcd.io/prune: disabled` annotation (GitOps), and the alchemy declaration carries a new `retain: true` prop (persisted to output). `deleteKroResource` honors `retain` by skipping the delete and dropping only the state entry — mirroring the existing deferred shared-RGD delete. Two factories on the same workload namespace resolve the same namespace name and the same retained declaration id (no conflicting owned copies).
*Tests:* `bootstrap-instance-namespace.test.ts` — "marked retained (prune-disabled) shared infrastructure", "alchemy declaration is retained (never deleted on teardown)", and "two factories on the same workload namespace produce the same retained control-plane namespace".

**P2 — derived name could exceed the 63-char namespace limit.**
`controlPlaneNamespaceFor()` now returns `<name>-kro` verbatim when it fits, otherwise truncates a prefix of the workload name and appends a short deterministic sha256 hash of the *full* name before the suffix — always ≤63 chars, stable, DNS-1123-valid, and distinct for distinct inputs (the hash disambiguates two long names sharing a truncated prefix).
*Tests:* new `test/core/control-plane-namespace.test.ts` (short passthrough, ≤63 for a 60+ char input, determinism, distinct-for-distinct, DNS-1123 validity).

**P3 — docs drift.**
`docs/advanced/alchemy-integration.md` now documents `ownsInstanceNamespace`, the extra retained control-plane Namespace declaration (three declarations instead of two), spec-derived placement, and the retain/prune-disable behavior. `docs/api/rook/index.md` replaces the stale "create the factory's control-plane namespace before deploying" guidance with the auto-derived + retained behavior (the other bootstrap API docs were already accurate).

**Guard intact:** a composition that self-owns its instance namespace without `ownsInstanceNamespace` (or with an explicit `instanceNamespace` equal to an owned namespace) still throws `UNSAFE_KRO_NAMESPACE_OWNERSHIP`.

**Verification:** `bunx tsc --noEmit` clean; `bun run lint` clean. Green: bootstrap-instance-namespace + kro-instance-safety + new control-plane-namespace tests (29), affected factory suites dagster/clickstack/clickhouse/valkey/rook (229), serialization + kro-factory + alchemy suites (384). Two clickstack/clickhouse `factory-modes` tests were updated to use the explicit `instanceNamespace` option — they previously asserted the CR namespace via a substring that only matched because the old baked value (`typekro-system-kro`) *contained* `typekro-system`; the corrected P1-a behavior derives placement from `spec.namespace`. `test/core` shows the same 9 pre-existing CEL member-expression / dependency-extraction failures as master — zero new failures.

yehudacohen:
## Reworked: pivot to a single control-plane namespace + auto-detection

This revision replaces the per-workload `<ns>-kro` + `ownsInstanceNamespace`-flag approach with a **single, stable control-plane namespace and automatic detection**. That collapses the three round-2 findings structurally rather than patching each one.

### The new model
- **One constant control-plane namespace: `typekro-system`** (`KRO_INSTANCE_CONTROL_PLANE_NAMESPACE`), consistent with the existing `typekro-singletons`. Every relocated instance CR lands here; override per factory with the existing `instanceNamespace` option. Because it's a constant it's always a valid DNS-1123 label — the `controlPlaneNamespaceFor()` derivation, hash-truncation, and length constants are gone (and their test removed).
- **Auto-detect self-ownership, no flag.** The guard's ownership detection is extracted into `detectOwnedNamespaces()` and reused: when a composition owns the namespace its instance would otherwise land in, the CR is **auto-relocated** to `typekro-system` (with a one-line diagnostic) instead of throwing. The `ownsInstanceNamespace` flag is **removed** from the RGD/factory types and all 7 bootstraps.

### How each round-2 finding is resolved
- **[P1] getInstances/deleteInstance orphan.** Both now **discover by RGD label cluster-wide** rather than a spec-derived namespace, so they find the CR wherever it landed. `deleteInstance` targets the discovered namespace and **rejects ambiguous same-name matches** instead of treating a wrong-namespace 404 as success — no more orphaned CRs/workloads. A stable constant also means a spec-less lifecycle call resolves the *same* namespace the deploy used.
- **[P1] Flux-only retention.** The control-plane namespace is emitted as a **deduped singleton** (declaration id keyed to the namespace name → one retained state entry across all consumers) annotated for **both** reconcilers — `kustomize.toolkit.fluxcd.io/prune: disabled` (Flux) **and** `argocd.argoproj.io/sync-options: Prune=false` (Argo) — plus the alchemy `retain` prop. Docs note it can be pre-created out-of-band for other GitOps tools.
- **[P2] Unmarked namespace-owning built-ins.** Auto-detection makes coverage uniform: the 7 former-flag bootstraps **plus** cert-manager / CNPG / caddy / apisix / external-dns / inngest / searxng all relocate automatically — nothing left throwing. A parametric test exercises all seven.

### Guard protection intact
The guard still throws `UNSAFE_KRO_NAMESPACE_OWNERSHIP` when a user **explicitly pins** `instanceNamespace` to a namespace the composition owns (opting back into the unsafe pattern), and fails closed on an owned Namespace whose name can't be evaluated. Covered by tests, including through composition nesting.

### Round-1 findings
Still addressed: the regression that broke typekro's own bootstraps is fixed (they serialize/relocate instead of throwing), and the docs (`alchemy-integration` + per-bootstrap API pages) are updated to the new model.

`bunx tsc --noEmit` and `bun run lint` (biome) are clean. Affected suites green (kro-instance-safety, bootstrap-instance-namespace, kro-factory, alchemy-deployers, rook/valkey/dagster/clickstack/clickhouse factory modes, new lifecycle+built-in coverage); the only remaining `test/core`/`test/unit` failures are the pre-existing strict-CEL member-expression/dependency-extraction ones, unrelated to this change.

yehudacohen:
Merged `origin/master` (v0.26.0) into this branch.

**Merge result:** clean, zero conflicts. The two sides touched completely disjoint files, so there was nothing to hand-resolve and no risk of silent semantic loss:
- **This branch** (the self-owned-namespace redesign): `kro-factory.ts`, `kro-instance-safety.ts`, alchemy `resource-registration.ts`/`types.ts`, `config/defaults.ts`, `types/deployment.ts` + tests.
- **Master** (NATS/JetStream PR #112 + deployment/teardown hardening): `direct-factory.ts`, `engine.ts`, `readiness-waiter.ts`, `rollback-manager.ts`, `src/factories/nats/**` + tests.

Master's "harden update readiness and teardown" work landed in the direct-factory/engine/readiness-waiter/rollback-manager path rather than `kro-factory.ts`, so the anticipated `kro-factory.ts` conflict did not materialize. Both the namespace/lifecycle redesign and master's hardening are fully preserved.

**NATS brought under the redesign:** NATS owns its workload Namespace, so auto-detection relocates its KRO instance CR to the single control-plane namespace `typekro-system` with **no nats-specific code change** — verified via the emitted YAML (`kind: NatsBootstrap` at `namespace: typekro-system`, plus the retained singleton Namespace with the `typekro.io/kro-instance-namespace` label and the Flux/Argo prune-disable annotations).

**Test brought in line:** `test/factories/nats/factory-modes.test.ts` previously asserted the old `.toThrow('cannot also be an owned Namespace')` for the same-namespace owner call — the single CI failure when the PR is merged into master. Reworked it to mirror the already-updated valkey factory-modes test: assert auto-relocation (instance lands in `typekro-system`, no throw), keep a paired assertion that an **explicit** `instanceNamespace` pin to an owned namespace **still throws**, and keep the `namespaceOwnership: external` no-throw case. `test/integration/nats/**` uses dynamic namespaces + cluster-wide `deleteInstance()` discovery, so it's already compatible — no change needed.

**Verification (local, `TYPEKRO_STRICT_CEL=1`):**
- Previously-failing `test/factories/nats/factory-modes.test.ts`: reproduced the failure pre-fix, now **5/5 pass**.
- kro-instance-safety + kro-instance-namespace-lifecycle + bootstrap-instance-namespace + alchemy-deployers + readiness-waiter + rollback-manager + deployment-readiness: **131/131 pass**.
- Broad `test/factories`: **733 pass, 0 fail**.
- `test/core` + `test/unit`: 4013 pass, **10 fail — identical to the 10 pre-existing strict-CEL analyzer failures on clean `origin/master`** (verified against a fresh v0.26.0 checkout); zero new failures.
- `bunx tsc --noEmit -p tsconfig.json`: clean. `bun run lint` (biome, 839 files): clean.

yehudacohen:
## Pivoted: retained workload namespace, not instance relocation

Reworked this PR from **relocating** the KRO instance CR to a single control-plane namespace (`typekro-system`) to **leaving the instance in its natural namespace and hoisting the owned Namespace out of the RGD graph**.

The guard's only real concern is that the instance's namespace is a graph-**owned child** (KRO GCs children before clearing the owner CR's finalizer → stranding on delete). So rather than moving the instance, we make the namespace *not a graph child*: it's removed from the RGD's `spec.resources` and emitted as a **retained** resource created deps-first (outside the graph). Because it's no longer a graph child, deleting the instance can never GC it → no finalizer stranding → the guard passes with the instance in its natural namespace.

### Mechanism
- `resolveInstanceNamespace(spec)` = explicit `instanceNamespace` ?? `spec.namespace ?? factory namespace`. No relocation; `KRO_INSTANCE_CONTROL_PLANE_NAMESPACE` and the single-control-plane idea are gone.
- Hoisting reuses the guard's detection (`detectOwnedNamespaceResources`) to resolve schema-driven names concretely, matches ids back to `this.resources`, and removes the owned Namespace whose name equals the **candidate CR namespace**. Emitted retained + deduped by namespace **name** (shared namespace → one declaration). Both the factory RGD (`factory('kro').toYaml()`) and the graph-level RGD (`graph.toYaml()`) hoist, so they stay consistent.
- Retention: `kustomize.toolkit.fluxcd.io/prune: disabled` (Flux) + `argocd.argoproj.io/sync-options: Prune=false,Delete=false` (Argo) + alchemy `retain: true`. Deps-first: `Namespace → RGD → instance`.

### The 6 review findings, each resolved structurally
1. **Shared namespace collapses distinct instances** — the instance stays in its natural namespace and its alchemy id is qualified by that namespace, so `analytics/dev` and `analytics/prod` are distinct resources; the second never clobbers the first. (Reproduced the collapse on the old code first — both → `dagsterbootstrapAnalytics` in `typekro-system`; new code → `dagsterbootstrapAnalyticsDev` / `…Prod` in `dev`/`prod`.) Regression test added.
2. **Delete fails-open on RBAC/transient errors (orphan)** — removed the cluster-wide `listClusterCustomObject` label-discovery entirely; `getInstances`/`deleteInstance` use the factory namespace (pre-guard behavior). The cleanup RGD-preservation list already fails **closed** (preserves the RGD when the list errors).
3. **Already-safe placements needlessly migrated** — detection compares the owned name against the **candidate CR namespace**, not blindly against `spec.namespace`; a factory whose instance namespace differs from the owned namespace is left untouched. Test added.
4. **Argo `Prune=false` doesn't survive Application deletion** — now `Prune=false,Delete=false` (+ alchemy `retain`), which survives an Argo Application deletion, not just a sync-prune.
5. **Order-dependent detection cache** — the cache is gone; the owns-namespace decision is computed per-spec at deploy/toYaml/toAlchemyResources time. Order-independence test added (conditional namespace false-then-true == true-first).
6. **Discovered namespace discarded** — no discovery needed; the instance stays in its natural namespace and lifecycle ops target the factory namespace directly.

Guard still throws for a genuinely unprovable owned Namespace name (fails closed) and for an explicit `instanceNamespace` pinned onto an owned namespace.

### Verification
`bunx tsc --noEmit` clean; `bun run lint` clean. Rewritten suites (safety, lifecycle, bootstrap-instance-namespace, dagster/clickhouse/clickstack/valkey/rook/nats factory-modes, characterization) + serialization all green. `test/core` (9) and `test/unit` (1) show the same pre-existing strict-CEL-analyzer + flaky-timeout failures as clean `v0.26.0` master — **zero new failures** (diff of failure sets is empty). Deps-first ordering verified (Namespace → RGD → instance).

PR #114, opened 2026-07-17
fix(factories): declare Kubernetes-integral schema fields as `number.integer` (strict-CEL)

yehudacohen:
## Problem

KRO's strict CEL type-checker (0.25+) **rejects** a `float` (double) schema field used where an int is expected. A composition that types an integer field as bare arktype `number` — e.g. a port at `spec.template.spec.containers[].ports[].containerPort` — fails at deploy:

```
expression "schema.spec.httpPort" returns type "double" but expected "int"
```

## Fix (revised per review)

arktype's bare `number` **correctly** maps to KRO `float` — it admits fractional values — so `getKroTypeFromJson` is left unchanged. Integrality must be **declared, not inferred**. This migrates every Kubernetes-integral factory field (ports, replicas, counts, `servicePort`, `containerPort`, `*Seconds`, thresholds, shards, DB numbers, status counts, …) from bare `number` to `number.integer` across all 13 factories — 128 fields.

- The one genuinely fractional field, rook `targetSizeRatio`, stays `number`.
- Float remains fully expressible (bare `number`, or fractional bounds/divisor).
- Adds a factory-level regression asserting caddy ports serialize as KRO `integer`.

## Supersedes the earlier heuristic

The first revision inferred integrality from the *absence* of fractional constraints. That was unsound, as the review correctly identified:
- **[P1] absence of fractional constraints ≠ integer** — it broke rook `targetSizeRatio` (a real `number` float) and made unrestricted floats inexpressible.
- **[P1] contradictory defaults** — `ratio?: number` with `?? 0.5` produced `integer | default=0.5`, invalid SimpleSchema.
- **[P2] exclusive fractional bounds missed** — `number > 0.1` is `min: { exclusive, rule }`, not a bare number, so the check emitted `integer`.

Reverting `schema.ts` to its `number → float` mapping and migrating the factory fields resolves all three: genuine floats stay float, defaults stay consistent, and no bound inspection is involved.

## Validation
- Full suite: **4766 pass / 0 fail** (269 files, via the pre-push hook).
- Lib typecheck passes.


yehudacohen:
Thanks — all three findings were valid; I've reworked the PR.

Confirmed `rook targetSizeRatio` is a genuine `number` float (it's the counterexample that sinks inferring integrality from missing evidence), reproduced the contradictory `integer | default=0.5`, and confirmed the exclusive-bound shape (`min: { exclusive, rule }`) was missed.

New approach follows your recommendation: **keep bare `number` → `float` (`schema.ts` reverted to unchanged) and migrate the Kubernetes-integral factory fields to `number.integer`** — 128 fields across all 13 factories, with `targetSizeRatio` left as `number`. Added a caddy factory regression that ports serialize as KRO `integer`. Full suite green (4766/0).

yehudacohen:
Good catches — both fixed in the follow-up (155db75):

- **apisix stream ports** `'number[]'` → `'number.integer[]'` (tcp/udp): now serialize as KRO `[]integer` (verified), rejecting `[9000.5]`.
- **ory int-or-string fields** (nodePort, rollingUpdate maxSurge/maxUnavailable, PDB minAvailable/maxUnavailable) `'string | number'` → `'string | number.integer'`. Note KRO SimpleSchema collapses the union to `object`, so the integrality is enforced at the arktype validation layer — confirmed it now rejects `nodePort: 30000.5` / `maxSurge: 0.5` while still accepting integers and strings.

Swept all single-quoted arktype schema strings for remaining bare `number` in array/union/record forms; the only other matches are `typeof x === 'number'` runtime guards and rook `targetSizeRatio` (the genuine float, left as `number`). Extended the regression test to cover the array serialization + int-or-string fractional rejection. Full suite 4768/0.

PR #116, opened 2026-07-20
fix(readiness): bound per-call k8s reads so a wedged exec credential can't hang the poll forever

yehudacohen:
## Problem

Readiness polls (`waitForKroInstanceReady`, `ResourceReadinessChecker.waitForResourceReadyWithPolling`) loop as:

```ts
while (Date.now() - startTime < timeout) { await k8sApi.read(...) }
```

The deadline is only re-evaluated **between** iterations, so it relies on each `await` eventually settling. When the kubeconfig's exec credential **wedges** — e.g. an AWS SSO / EKS `get-token` credential that expired mid-deploy so the auth plugin never returns — the awaited request never resolves **or** rejects. The loop never re-checks its deadline and the deploy hangs **silently** (observed multi-hour; the configured `timeout` is effectively ignored on the credential-failure path).

## Fix

New `callWithTimeout()` (`src/core/deployment/poll-timeout.ts`) bounds each readiness API call to `DEFAULT_HTTP_READ_TIMEOUT`, capped **strictly** by the poll's remaining deadline (`perCallTimeout`) so one call can't overshoot the overall `timeout`. A wedge becomes a **rejection**, so the existing error handling runs and the deadline is honored.

Distinct failure modes are kept distinct:
- **A call that actually started and timed out** → `PollTimeoutError` (credential-wedge hint). In `waitForKroInstanceReady` the outer catch re-throws it (fail fast); in `waitForResourceReadyWithPolling` it's logged and retried within the deadline.
- **The overall deadline already elapsed** (per-call budget ≤ 0) → the caller **breaks** and throws its own overall-timeout error (`DeploymentTimeoutError` / `ResourceReadinessTimeoutError`). These are never conflated.

Related correctness:
- The RGD status-schema read is bounded too, and its `PollTimeoutError` is **re-thrown** rather than swallowed by the permissive 'RGD unfetchable' path (which would declare an ACTIVE/synced instance ready without validating expected status fields). A deadline re-check guards the ready-return.
- The retry sleep and the on-timeout debug log never run past the deadline: the sleep is capped to the remaining budget, and the debug log reuses the **last observed** resource instead of issuing a fresh post-deadline read.

## Scope / limitation (finding acknowledged)
This bounds the caller's `await` (poll + deploy terminate with an actionable error). It does **not** cancel the in-flight request or kill a wedged exec-auth subprocess: `@kubernetes/client-node` 1.3.0's `read()` doesn't thread an `AbortSignal` to fetch and `ExecAuth` spawns the credential via `child_process.spawn` with no cancellation hook. That's documented in `poll-timeout.ts`; the durable cure for the exec-wedge is avoiding per-request exec auth during polling (a pre-minted bearer token in the kubeconfig).

## Tests
- `test/unit/poll-timeout.test.ts`: never-settles → `PollTimeoutError` with credential hint; real rejections propagate; timer cleared; `perCallTimeout` caps strictly to remaining (no floor) and returns ≤0 when spent.
- `test/unit/kro-readiness.test.ts`: wedged RGD read fails fast (not false-ready); an exhausted deadline throws `DeploymentTimeoutError`, not `PollTimeoutError`.
- Full unit suite: **4850 pass, 0 fail**; `tsc` + `biome` clean.

### DIFF
--- a/src/core/deployment/kro-readiness.ts
+++ b/src/core/deployment/kro-readiness.ts
@@ -19,7 +19,7 @@ import {
 import { CRDInstanceError, DeploymentTimeoutError, ensureError } from '../errors.js';
 import { getComponentLogger } from '../logging/index.js';
 import type { RGDManifest } from '../types/kubernetes.js';
-import { callWithTimeout, perCallTimeout } from './poll-timeout.js';
+import { callWithTimeout, perCallTimeout, PollTimeoutError } from './poll-timeout.js';
 
 /** Options for Kro instance readiness polling. */
 export interface KroReadinessOptions {
@@ -184,6 +184,13 @@ export async function waitForKroInstanceReady(options: KroReadinessOptions): Pro
           expectedCustomStatusFields,
         });
       } catch (error: unknown) {
+        // A per-call TIMEOUT (wedged/expired credential) is NOT a fetchable-RGD failure — do not fall
+        // through to the permissive path, which would let an ACTIVE/synced instance be declared ready
+        // WITHOUT validating expected status fields (and after the deadline). Surface it so the outer
+        // catch re-throws it (fail fast).
+        if (error instanceof PollTimeoutError) {
+          throw error;
+        }
         readinessLogger.warn('Could not fetch ResourceGraphDefinition for status schema check', {
           rgdName,
           error: ensureError(error).message,
@@ -232,6 +239,12 @@ export async function waitForKroInstanceReady(options: KroReadinessOptions): Pro
         (!hasReadyField || statusReadyField === true);
 
       if (isReady) {
+        // Re-check the deadline: an API call earlier in THIS iteration may have consumed the remaining
+        // budget (e.g. a slow/bounded read), so an "isReady" reached past the deadline must still time
+        // out rather than silently succeed late.
+        if (Date.now() - startTime >= timeout) {
+          break;
+        }
         readinessLogger.info('Kro instance is ready', {
           instanceName,
           hasCustomStatusFields,
--- a/src/core/deployment/kro-readiness.ts
+++ b/src/core/deployment/kro-readiness.ts
@@ -11,10 +11,15 @@
  */
 
 import type * as k8s from '@kubernetes/client-node';
-import { DEFAULT_FAST_POLL_INTERVAL, DEFAULT_POLL_INTERVAL } from '../config/defaults.js';
+import {
+  DEFAULT_FAST_POLL_INTERVAL,
+  DEFAULT_HTTP_READ_TIMEOUT,
+  DEFAULT_POLL_INTERVAL,
+} from '../config/defaults.js';
 import { CRDInstanceError, DeploymentTimeoutError, ensureError } from '../errors.js';
 import { getComponentLogger } from '../logging/index.js';
 import type { RGDManifest } from '../types/kubernetes.js';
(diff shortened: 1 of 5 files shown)


yehudacohen:
Thanks — all four are addressed in `83155db`.

**[P1] RGD timeout swallowed as readiness success** — Fixed. Added a distinguishable `PollTimeoutError`; the RGD-fetch `catch` now re-throws it (instead of falling through to the permissive `expectedCustomStatusFields = false` path), and the outer catch re-throws it → fail fast. Also added a deadline re-check before returning ready, so a late 'ready' still times out. Regression test: a wedged RGD read on a 200ms deadline now **rejects** instead of returning ready (previously returned ready ~1s in).

**[P1] Final debug read can still hang** — Fixed. The post-deadline debug-status read is now wrapped in `callWithTimeout`, so a wedged credential surfaces `ResourceReadinessTimeoutError` instead of hanging there.

**[P2] One-second floor violates remaining deadline** — Fixed. `perCallTimeout` now caps **strictly** to the remaining budget (no 1s floor) and returns ≤0 once spent; `callWithTimeout` fails fast on a ≤0 budget without starting a doomed call. (This was the other half of the 1008ms-on-100ms repro.)

**[P1] Promise.race leaves the exec credential running** — You're right, and I want to be precise about scope. Through `@kubernetes/client-node` 1.3.0's public API this isn't reachable from the poll layer: `KubernetesObjectApi.read` doesn't thread an `AbortSignal` to its fetch, and `ExecAuth` spawns the credential via `child_process.spawn` with no cancellation hook. So `Promise.race` bounds the **logical** wait (the poll + deploy terminate and surface an actionable error — the multi-hour-hang fix), but it cannot force-kill a wedged exec subprocess, which can outlive the rejection until it exits or the process is terminated. I've **documented this limitation explicitly** in `poll-timeout.ts` rather than implying cancellation it can't deliver.

The durable cure for that failure mode is to not spawn per-request exec auth during polling at all — a **pre-minted bearer token** in the kubeconfig. That's a client-construction change (and downstream I can do it in the foundry's `writeKubeconfig`); happy to add an option here too (e.g. an exec-auth timeout / token-refresh seam) if you'd prefer it live in typekro.

Full unit suite: **4850 pass, 0 fail**; `tsc` + `biome` clean.

yehudacohen:
Round 2 addressed in `4da8ae3`:

**[P1] Exhausted budget is not a credential timeout** — Fixed. `callWithTimeout` no longer throws on a ≤0 budget. Callers now guard: when `perCallTimeout` returns ≤0 (overall deadline elapsed) they **break** and throw their normal overall-timeout error (`DeploymentTimeoutError` / `ResourceReadinessTimeoutError`). `PollTimeoutError` is reserved for a call that actually started and timed out. New regression test asserts a 1ms deadline → `DeploymentTimeoutError` (and the message is *not* the credential hint). This also fixes the CI failure.

**[P2] Retry sleep runs after deadline** — Fixed. The `errorRetryDelay` sleep is capped to the remaining budget (0 → skipped), so a bounded read that consumed the budget can't overshoot (no more 2106ms-on-100ms).

**[P2] Debug read receives a fresh 30s budget** — Fixed. The on-timeout debug log now uses the **last observed** resource from the loop; no fresh post-deadline read (so it can't add 30s or hang).

**Docs** — PR description updated: removed the 1s-floor mention and refreshed the test counts.

Full unit suite: **4850 pass, 0 fail**; `tsc` + `biome` clean.

PR #122, opened 2026-07-27
ci: report every gate instead of stopping at the first failure

yehudacohen:
A failing step skips every later step, so a single red gate hides the rest of the signal. This bit us concretely: the **Effect dependency cohort** gate failed and masked a **failing packed-artifact budget** behind it — CI never even reached the packaging gate, so the budget failure was invisible until run locally. Each fix/push cycle then only reveals the next problem.

## Change
Add `if: ${{ !cancelled() }}` to the verification gates after `Build`, so all of them run and report in one pass:

Typecheck · Lint · Docs validation · Circular dependency check · Effect dependency cohort · Packed artifact budgets · Randomized unit test isolation · Unit tests with coverage · Integration YAML tests

Semantics are unchanged otherwise:
- the job still **fails** if any gate fails (each step's own status stands),
- `Build` stays a hard gate — packaging and tests need `dist/`, so continuing past a broken build produces noise, not signal,
- the Codecov upload keeps `if: success()`, so it still only uploads on a fully green run.

Net effect: one CI run tells you everything that's broken instead of one thing at a time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,12 +26,15 @@ jobs:
         run: bun run build
 
       - name: Typecheck
+        if: ${{ !cancelled() }}
         run: bun run typecheck
 
       - name: Lint
+        if: ${{ !cancelled() }}
         run: bun run lint
 
       - name: Docs validation
+        if: ${{ !cancelled() }}
         run: |
           bun run docs:validate
           bun run docs:check-lines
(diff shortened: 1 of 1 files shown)


yehudacohen:
Good catch — fixed in `023879b`, and I then exercised both failure paths to close the "only the success path is tested" gap.

## Fix
`if: ${{ !cancelled() }}` does drop the implicit `success()`, so the gates were also running after a failed checkout / install / Build. That's worse here than in general, because `bun run build` is `clean && lint && typecheck && build:lib && build:examples` — a lint failure wipes `dist/` and then cascades into bogus packaging and test failures.

Applied your recommendation: `id: build` on Build, and `if: ${{ !cancelled() && steps.build.outcome == 'success' }}` on the nine verification gates. A *skipped* Build (from a failed checkout or install) reports `skipped`, so those cases short-circuit too.

## Verified on a throwaway PR (#123, now closed, branch deleted)

**Probe 1 — Build fails** (build script forced to `exit 1`):
```
success  Install dependencies
failure  Build
skipped  Typecheck / Lint / Docs validation / Circular dependency check
skipped  Effect dependency cohort / Packed artifact budgets / Randomized / Coverage / Integration YAML
```
No gate ran against a broken build; job correctly red.

**Probe 2 — Build green, one mid-list gate fails** (cohort gate pinned to a bogus version — the exact scenario that motivated this PR):
```
success  Build / Typecheck / Lint / Docs validation / Circular dependency check
failure  Effect dependency cohort
success  Packed artifact budgets and imports     <-- previously skipped/hidden
success  Randomized unit test isolation
success  Unit tests with coverage
success  Integration YAML tests
skipped  Upload coverage to Codecov
```
The four gates after the failure now run and report, which is the whole point — previously the cohort failure hid the packaging gate entirely. Codecov still correctly skips on a non-green run, and the job is still red.

Incidentally, `Packed artifact budgets and imports` passes on the runner. I'd measured the Node cold-import time straddling its 5s budget locally (11.2s / 5.3s / 3.2s across three runs), so that flakiness may be local-only — worth keeping an eye on rather than acting on now.

PR #123, opened 2026-07-27
[TEMP PROBE — do not merge] verify CI gates skip when Build fails

yehudacohen:
Throwaway draft to exercise the failure path for #122 (the review noted CI only proves the success path). Build is forced to exit 1; expectation is that all 9 verification gates report **skipped**, not failed. Will be closed and the branch deleted as soon as the run reports.


yehudacohen:
Probe complete — both paths verified, see evidence on #122. Closing; never intended to merge.

PR #130, opened 2026-07-29
fix(core): map arktype `object` primitive to KRO `object` (fixes #128)

### LINKED ISSUE #128: dagster: postgresql subchart values are unreachable through the KRO path (spec.values.postgresql is schema-constrained; spec.postgresql.values is typed string)
Setting any bundled-**postgresql subchart** value through `dagsterBootstrap.factory('kro')` is silently dropped. Both candidate seams fail, in different ways.

## Seam 1 — `config.values.postgresql` (raw chart values): schema-constrained, so pruned

`DagsterHelmValues.postgresql` is declared `TypeKroValueTreeObject` (free-form), but the generated RGD schema constrains it to the *convenience* field set:

```yaml
postgresql:
  authProvider: string
  database: string
  enabled: boolean
  host: string
  params: map[string]string
  password: string
  passwordSecretName: string
  scheme: string
  servicePort: integer
  storageClass: string
  username: string
  values: string
```

`primary`, `nodeSelector` and `persistence` aren't there, so KRO admission strips them. Confirmed live:

```
# we sent values.postgresql.primary.persistence.enabled=false and .persistence.enabled=false
kubectl get dagsterbootstrap -n <ns> -o jsonpath='{.items[0].spec.values.postgresql}'
=> {"enabled":true}
```

## Seam 2 — `config.postgresql.values` (the documented escape hatch): typed `string`

`DagsterPostgresqlConfig.values` is `TypeKroChartValue<TypeKroValueTreeObject>` and documented as *"Raw official postgresql subchart values merged before top-level `values`"* — exactly the right seam. It **is** consumed in CEL (`json.unmarshal(json.marshal(schema.spec.postgresql.values))`, merged into the release values), but the RGD simple-schema types it `string`, and passing an object is rejected 422.

So the declared TS type (object tree) and the generated schema type (`string`) disagree, and the field ends up unusable either way.

## Impact

`persistBundledPostgres: false`-style config reads as applied and isn't — the accepted-but-ignored-config failure mode. In our case per-PR ephemeral environments were meant to run Postgres on ephemeral storage; they each get an 8Gi PVC instead, and the lingering PVC also keeps the namespace non-empty, defeating an empty-gated namespace teardown.

Same class as the `values.global.env` pruning: unit tests and serialization goldens both pass, because the golden happily contains the field that admission later strips. **Verify any fix by reading the CR back after admission**, not from the golden.

## Suggested fix

Make the subchart passthrough genuinely opaque — the RGD schema for `postgresql.values` (and by symmetry `rabbitmq.values` / `redis.values`, which look identical) should preserve unknown fields rather than being typed `string`, so the existing CEL merge receives a real map.

Happy to send a PR if you can confirm the intended direction — whether the escape hatch should be preserve-unknown-fields, or `DagsterHelmValues.postgresql` should stop being schema-narrowed.

yehudacohen:
Fixes #128, addressing the review's recommendation to fix this in **core** rather than applying a Dagster workaround.

## Root cause (P1, as the review identified)

Arktype represents a bare `type('object')` as the **string** `"object"` — not as `{ domain: "object" }`, which the converter's object branch already handled. The primitive switch had no case for it, so it fell through to the `'string'` default. Every schemaless-object field was therefore declared `string` in the generated RGD, and KRO admission dropped whatever object a caller sent.

`object[]` needed no separate case: the array branch recurses through the same switch, so a `sequence` of `"object"` now yields `[]object`.

**Blast radius in the Dagster RGD alone:** `object` **0 → 36** fields, `[]object` **22 → 38**. Confirmed core-wide, not Dagster-specific — APISIX, Ory and ClickHouse use the same construct.

## Dagster escape hatch (P1)

`values` was assigned the fully typed `helmValuesSchemaShape`, making a **closed** schema out of a field advertised as *"raw official chart values"*. Worse, its `postgresql` sub-shape reused the TypeKro **convenience** shape, so bundled-subchart settings (`primary.*`, `nodeSelector`, `persistence`) were unreachable through *either* documented route.

`values` is now opaque. Typed convenience stays on the high-level API (`postgresql`, `webserver`, `runLauncher`, …), and the `DagsterHelmValues` TS interface still documents the common fields — which is exactly the split the review recommended. The now-dead `helmValuesSchemaShape` is removed, with a comment in its place so nobody re-wires it.

One find worth calling out: the `serviceAccount.annotations` comment asserted that `object` *"serializes to `type: string`"* as if that were expected behaviour. That was this bug, documented as a feature. `Record<string,string>` remains correct there (annotations really are a string→string map), but the rationale is fixed.

## Tests (P2)

The review was right that the old tests proved references, not behaviour. Added:

- **Core:** `object` → `object`, `object[]` → `[]object`, required-vs-optional, and the structured forms that must **not** change (`Record<string,string>` stays `map[string]string`; nested shapes stay nested).
- **RGD types:** exact assertions for `spec.values`, `spec.postgresql.values`, `spec.rabbitmq.values`, `spec.redis.values` — plus a guard that opaque `values` didn't swallow the typed API.
- **Live KRO admission** (`test/integration/dagster/values-admission.test.ts`): applies the RGD, submits a CR with nested unknown maps/booleans/numbers/arrays, and reads the admitted object back. Image-free, so it needs only a cluster with KRO.

**I verified the live test fails without the core fix** — a test that can't fail proves nothing.

## Why this stayed hidden

The defect surfaces differently per client. `kubectl apply` uses strict decoding, so it errors loudly (`strict decoding error: unknown field "spec.values.postgresql"`) — that's the "422/BadRequest" people hit poking by hand. TypeKro's own apply path is **not** strict, so in production the identical mismatch was **pruned in silence** and the converge reported success. The new test drives kubectl deliberately, so a regression is a hard error.

## Verification

Typecheck clean. Full unit suite **5158 pass / 0 fail** — notable given 36+38 field types changed. Live admission test 3/3 against orbstack.

## Not done here

The review's audit item: APISIX/Ory/ClickHouse benefit automatically from the core fix (their `type('object')` fields now map correctly, and the core tests cover the construct), but I did not add per-factory live admission tests for them. Worth a follow-up if you want the same guarantee everywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,55 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Arktype `object` now maps to KRO `object` instead of `string`.** A bare `type('object')` is represented by
+  arktype as the string `"object"`, which the KRO type mapper had no case for, so it fell through to the
+  `string` default. Every schemaless-object field was therefore declared `string` in the generated RGD and KRO
+  admission dropped whatever object a caller sent. `object[]` is fixed by the same change (the array branch
+  recurses). Affects any factory using `type('object')` for passthrough config — Dagster, APISIX, Ory,
+  ClickHouse — not just Dagster.
+- **Dagster's raw `values` escape hatch is genuinely opaque.** It had been assigned the fully typed
+  `helmValuesSchemaShape`, making a CLOSED schema out of a field documented as *raw official chart values*; its
+  `postgresql` sub-shape reused the convenience shape, so bundled-subchart settings (`primary.*`,
+  `nodeSelector`, `persistence`) were unreachable through either documented route. Typed convenience remains on
+  the high-level API (`postgresql`, `webserver`, `runLauncher`, …).
+
+### Added
+
+- **`allowBreakingChanges` factory option.** Stamps `kro.run/allow-breaking-changes: "true"` on the generated
+  RGD. The same option already existed at composition level, but a consumer of a SHIPPED composition
+  (`dagsterBootstrap`, `apisixBootstrap`, …) cannot reach that, so there was no way to migrate an
+  already-deployed RGD. Off by default.
+
+### ⚠️ UPGRADE NOTE — existing deployments need one authorized converge
+
+The `object` correction changes the DECLARED TYPE of existing RGD fields (52 of them in the Dagster RGD alone).
+KRO refuses breaking CRD updates, **and the refusal is silent**: the apply succeeds, the RGD reports
+`Ready=True`, and the registered CRD keeps its OLD schema. The only evidence is a controller log line:
+
+```
+cannot update CRD <plural>.kro.run: breaking changes detected: Type changed from string to object; ...
+```
+
+So an upgrade can appear to land while nothing changed and values continue to be pruned. To migrate, authorize
+it for the one converge that performs the change:
+
+```ts
+myComposition.factory('kro', { allowBreakingChanges: true })
+```
+
+then drop the option again. Verify with:
+
+```
+kubectl get crd <plural>.kro.run \
+  -o jsonpath='{.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.values.type}'
+```
+
+It should print `object`. Deleting and recreating the RGD also works (KRO re-registers the CRD from scratch)
+but destroys existing custom resources, so prefer the annotation. Do not enable it globally: it disables a real
+safety check, and a genuinely lossy change (narrowing a type, dropping a field) can strand existing CRs.
+
 ## [0.21.0] - 2026-07-02
 
 ### Added
--- a/package.json
+++ b/package.json
@@ -145,7 +145,7 @@
     "test:packed-imports": "node scripts/test-packed-imports.mjs",
     "test:all": "bun run test && bun run test:integration:all",
     "test:integration": "./scripts/run-integration-tests.sh",
-    "test:integration:dagster": "bun test test/integration/dagster/bootstrap-composition.test.ts --timeout 900000",
+    "test:integration:dagster": "bun test test/integration/dagster/ --timeout 900000",
     "test:integration:required": "REQUIRE_CLUSTER_TESTS=true ./scripts/run-integration-tests.sh",
     "test:integration:debug": "DEBUG_MODE=true ./scripts/run-integration-tests.sh",
     "typecheck": "bun run typecheck:lib && bun run typecheck:examples && bun run typecheck:tests",
--- a/src/core/deployment/kro-factory.ts
+++ b/src/core/deployment/kro-factory.ts
@@ -4571,6 +4571,12 @@ export class KroResourceFactoryImpl<
       {
         ...(this.factoryOptions.compositionOptions as SerializationOptions | undefined),
         namespace: this.namespace,
+        // Factory-level override. The composition option of the same name is unreachable for a consumer of a
+        // SHIPPED composition (its definition lives in TypeKro), so without this there is no way to authorize
+        // the one-off CRD schema migration KRO would otherwise silently refuse.
+        ...(this.factoryOptions.allowBreakingChanges === undefined
+          ? {}
+          : { allowBreakingChanges: this.factoryOptions.allowBreakingChanges }),
       },
       kroSchema
     );
@@ -4600,10 +4606,12 @@ export class KroResourceFactoryImpl<
   }
 
   private buildRgdYaml(spec?: TSpec): string {
-    return serializeResourceGraphDefinitionToYaml(
-      this.buildRgdManifest(spec),
-      this.factoryOptions.compositionOptions as SerializationOptions | undefined
-    );
+    return serializeResourceGraphDefinitionToYaml(this.buildRgdManifest(spec), {
+      ...(this.factoryOptions.compositionOptions as SerializationOptions | undefined),
+      ...(this.factoryOptions.allowBreakingChanges === undefined
+        ? {}
+        : { allowBreakingChanges: this.factoryOptions.allowBreakingChanges }),
+    });
   }
 
   /**
--- a/src/core/serialization/schema.ts
+++ b/src/core/serialization/schema.ts
@@ -133,6 +133,18 @@ function getKroTypeFromJson(node: unknown): string {
       case 'string':
       case 'boolean':
         return node;
+      // Arktype represents a bare `type('object')` as the STRING "object" (not as
+      // `{ domain: "object" }`, which the object branch above already handles). Omitting it here fell
+      // through to the `'string'` default, so every schemaless-object field was declared `string` in the
+      // RGD — and KRO admission then PRUNED any real object a caller sent, silently. That is what made
+      // Dagster's documented Helm `values` escape hatches unusable, and it affects every factory using
+      // `type('object')` for passthrough config. KRO models `object` as schemaless (`type: object` with
+      // `x-kubernetes-preserve-unknown-fields: true`), which is exactly the representation needed.
+      //
+      // `object[]` needs no separate case: the array branch above recurses through here, so a
+      // `sequence` of "object" now yields `[]object`.
+      case 'object':
+        return 'object';
     }
   }
 
--- a/src/core/types/deployment.ts
+++ b/src/core/types/deployment.ts
@@ -670,6 +670,27 @@ export interface PublicFactoryOptions extends BaseDeploymentConfig {
    */
   instanceNamespace?: string;
 
+  /**
+   * Stamp `kro.run/allow-breaking-changes: "true"` on the generated ResourceGraphDefinition, authorizing KRO
+   * to apply a CRD schema change it would otherwise refuse.
+   *
+   * WHY THIS EXISTS AT FACTORY LEVEL. The same flag is available as a *composition* option, but a consumer of
+   * a SHIPPED composition (`dagsterBootstrap`, `apisixBootstrap`, …) cannot reach that — the composition is
+   * defined inside TypeKro. Without a factory-level switch there is no way to migrate an already-deployed RGD.
+   *
+   * WHEN YOU NEED IT. KRO refuses breaking CRD updates, and the refusal is EASY TO MISS: `kubectl apply`
+   * reports success, the RGD reports `Ready=True`, and only the controller log says
+   * `cannot update CRD ...: breaking changes detected`. The registered CRD silently keeps its old schema, so
+   * an upgrade appears to land while nothing changed. Set this for the one converge that performs the
+   * migration, then remove it.
+   *
+   * Scoped to a single factory deliberately — do not turn it on globally. It disables a real safety check, and
+   * a schema change that is genuinely lossy (narrowing a type, dropping a field) can strand existing CRs.
+   *
+   * @default false
+   */
+  allowBreakingChanges?: boolean;
+
   /** Explicit KubeConfig override for cluster connection */
   kubeConfig?: KubeConfig;
 
--- a/src/factories/dagster/types.ts
+++ b/src/factories/dagster/types.ts
@@ -418,36 +418,27 @@ const globalSchemaShape = {
 const serviceAccountSchemaShape = {
   'create?': 'boolean',
   'name?': 'string',
-  // Record<string,string> (not objectMapSchema/`object`) so the generated CRD field is an object map with
-  // string values — `object` serializes to `type: string`, which k8s rejects for an annotations map (422).
+  // `Record<string,string>` rather than `object`: annotations ARE a string→string map, so the precise type
+  // is the right one — it documents the contract and rejects a non-string value at the boundary.
+  // (The previous rationale here said `object` "serializes to `type: string`". That was true, but it was
+  // describing the converter BUG this workaround was hiding: arktype renders a bare `type('object')` as the
+  // string "object", which the KRO type mapper had no case for and defaulted to `string`. Fixed in
+  // core/serialization/schema.ts — `object` now maps to KRO's schemaless `object`.)
   'annotations?': 'Record<string, string>',
 } as const;
 
-const helmValuesSchemaShape = {
-  ...globalSchemaShape,
-  'global?': globalSchemaShape,
-  'serviceAccount?': serviceAccountSchemaShape,
-  'nameOverride?': 'string',
-  'fullnameOverride?': 'string',
-  'rbacEnabled?': 'boolean',
-  'imagePullSecrets?': namedRefSchema.array(),
-  'dagsterWebserver?': webserverSchemaShape,
-  'dagsterDaemon?': daemonSchemaShape,
-  'dagster-user-deployments?': userDeploymentsSchemaShape,
-  'postgresql?': postgresqlSchemaShape,
-  'generatePostgresqlPasswordSecret?': 'boolean',
-  'generateCeleryConfigSecret?': 'boolean',
-  'rabbitmq?': rabbitmqSchemaShape,
-  'redis?': redisSchemaShape,
-  'runLauncher?': runLauncherSchemaShape,
-  'scheduler?': schedulerSchemaShape,
-  'computeLogManager?': computeLogManagerSchemaShape,
-  'flower?': flowerSchemaShape,
-  'ingress?': ingressSchemaShape,
-  'pythonLogs?': objectMapSchema,
-  'busybox?': objectMapSchema,
-  'extraManifests?': 'unknown[]',
-} as const;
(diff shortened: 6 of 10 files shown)


yehudacohen:
Worked through the findings. **Finding 1 is confirmed and materially worse than described** — that part needs a decision from you before I write more code.

## Finding 1 — confirmed, and the failure is silent

I reproduced it end to end on KRO 0.9.2 rather than reasoning from the docs: applied the legacy (master) Dagster RGD, then applied this PR's corrected RGD.

`kubectl apply` reports success and **the RGD reports healthy**:

```
resourcegraphdefinition.kro.run/dagster-bootstrap configured
GraphAccepted=True Valid; KindReady=True Ready; ControllerReady=True Running; Ready=True Ready
```

But the registered CRD still has the old types:

```
values.type            = string
postgresql.values.type = string
```

The rejection only exists in the controller log, in a hot requeue loop:

```
Breaking changes detected in CRD update  breakingChanges: 52
  summary: "Type changed from string to object; ... and 49 more changes"
ERROR cannot update CRD dagsterbootstraps.kro.run: breaking changes detected
```

Your count of 52 is exact. So the upgrade hazard is **not** a loud failure at apply time — an operator upgrading TypeKro sees a successful apply, a Ready RGD, and a CRD that never changed. They'd conclude the fix landed while values continue to be pruned. Same silent-no-op signature as the original bug.

I also verified the remedy: applying the corrected RGD with `kro.run/allow-breaking-changes: "true"` updates the CRD (`values.type=object`).

**On the "no factory-level switch" point:** the plumbing half-exists. `allowBreakingChanges` is already a documented serialization option (`core/types/serialization.ts:313`) and is honoured in `yaml.ts:854`. What's missing is reachability — a `KroResourceFactory`'s `toYaml()` takes no arguments (`core/types/deployment.ts:927`), so a `kubernetesComposition` user genuinely cannot set it.

That makes the remaining work small but it's an API decision I'd rather you make: plumb `allowBreakingChanges` through `factory('kro', …)` options, or through `toYaml(options)`, or both. Say which and I'll add it with the legacy→corrected migration test.

## Findings 2 and 4 — fixed

- Removed `serviceAccountSchemaShape`, dead once `helmValuesSchemaShape` went. Worth noting its comment justified `Record<string,string>` for annotations because *"`object` serializes to `type: string`"* — this bug written up as a rule. The type was still right for annotations; the reason wasn't.
- Replaced `Array<any>` with a typed `RgdDocument` interface, as you preferred.
- `test:integration:dagster` now runs `test/integration/dagster/`, so the regression is included.

Lint and typecheck clean.

## Finding 3 — mostly done, one issue left

Rewritten to use `createTestNamespace` (UID lease), Bun-compatible k8s clients instead of shelling out to `kubectl`, tracking of every created object, and **aggregated** teardown failures rather than swallowed ones.

The substantive change: **it no longer creates any Dagster instance.** You were right that "image-free" wasn't a dependable property. Applying the Dagster RGD only registers a CRD, so that half asserts CRD types with no CRs at all; the schemaless round-trip runs against a minimal ConfigMap-only probe graph. That makes the property structural.

Still failing: the Dagster CRD assertion, and the cause is finding 1 biting the test itself — a CRD left over from an earlier run keeps its old types and KRO refuses the breaking update, so the assertion reads a stale schema. Deleting the RGD doesn't reliably reap the CRD. That wants the same migration mechanism, which is why I'd like your call on the switch first.

Also fixed along the way: `AlreadyExists` was being swallowed and the stale RGD reused, which would have silently asserted against the wrong definition.

Agreed this isn't mergeable yet.

yehudacohen:
All four findings addressed. **CI is green.** Finding 1 is confirmed and was worse than reported.

## 1 — existing RGDs cannot upgrade ✅

I reproduced it on KRO 0.9.2 rather than reasoning from the docs, and **KRO does not surface the refusal**:

```
resourcegraphdefinition.kro.run/dagster-bootstrap configured     # apply succeeds
GraphAccepted=True; KindReady=True; Ready=True Ready             # RGD looks healthy
values.type = string                                             # ...CRD never moved
```

The only evidence is a controller log line, in a requeue loop:

```
Breaking changes detected in CRD update  breakingChanges: 52
ERROR cannot update CRD dagsterbootstraps.kro.run: breaking changes detected
```

Your count of 52 is exact. So the hazard isn't a loud rejection — an operator sees a clean upgrade while nothing changed and values keep being pruned. Same silent-no-op shape as the bug being fixed.

**The switch:** `allowBreakingChanges` already existed at composition level and already reached the annotation, but a consumer of a shipped composition can't set composition options — `dagsterBootstrap`'s definition lives inside TypeKro. Threaded it through `PublicFactoryOptions`, so `factory('kro', { allowBreakingChanges: true })` works. Off by default, scoped per factory; I did **not** touch the global default.

**Tested migration path** (`schema-migration.test.ts`): seed a legacy `string` RGD, apply the corrected `object` schema **without** authorization and assert the CRD does *not* move while the RGD update is accepted — pinning the trap itself — then apply **with** authorization and assert it does. The legacy state is hand-authored so the before-state is explicit rather than dependent on git history.

CHANGELOG carries an upgrade note with the procedure and a `kubectl` verification command, plus why not to enable it globally.

## 2 — CI red ✅

Dead `serviceAccountSchemaShape` removed; `Array<any>` replaced with a typed interface. On the second pass I used a recursive `SchemaNode` type plus a small `at()` accessor, because `Record<string, unknown>` made every nested read `unknown` under the build's tsc.

Worth flagging: I had been running `bunx tsc --noEmit`, which is **not** what the repo checks — `typecheck` spans several project configs. That's why I shipped a red commit. Now validated with `bun run lint`, `bun run typecheck`, `bun tsc`, and `tsconfig.examples.json`.

## 3 — test could poison a cluster ✅

Harness `createTestNamespace` UID lease, Bun-compatible clients instead of `kubectl`, every created object tracked, aggregated teardown failures.

The substantive change: **it no longer creates any Dagster instance.** You were right that "image-free" wasn't dependable. Applying the RGD only registers a CRD, so that half asserts CRD types with no CRs; the round-trip uses a minimal ConfigMap-only probe graph. The property is now structural.

Your teardown point then bit me twice, both fixed:
- deleting a CRD while its CR's finalizer is pending leaves the namespace unable to terminate — the hook hung for the full 300s timeout. CRs must be gone *and settled* before their CRD goes.
- deleting an RGD and its CRD back-to-back lets KRO **re-register** the CRD, leaving one behind per run. Now waits for the RGD to be observed gone.

Verified: 3 consecutive runs, 4 pass / 0 fail, and zero leftover namespaces, RGDs or CRDs after a settle window.

## 4 — script omitted the regression ✅

`test:integration:dagster` now runs `test/integration/dagster/`.

Note `bootstrap-composition.test.ts` fails in that directory without `TYPEKRO_TEST_STORAGE_CLASS` — pre-existing and unrelated; the script already pointed at that exact file before this change.

## Verification

lint clean · typecheck 0 · `bun tsc` 0 · examples 0 · unit **5159 pass / 0 fail** · integration 4/4 live on orbstack, thrice.

yehudacohen:
All three findings addressed in `220a441` — thanks, the P1 was a live footgun.

## [P1] Shared Dagster definitions no longer deletable
`applyRgd` now takes an explicit ownership policy:

- **`'test-owned'`** (the probe) — keeps delete-and-recreate, because a leftover from an earlier run carries an old schema and reusing it would silently assert against the wrong definition.
- **`'shared'`** (the real `dagster-bootstrap` + its HelmRepository singleton) — **updates in place** via a `resourceVersion` compare-and-swap, and is **never registered for teardown**. The `'shared'` branch `continue`s before any delete or `created.push`, so there is no code path that can remove a production-named RGD.

The assertions are unchanged: KRO re-registers the CRD from the updated schema either way, so the live round trip still exercises the real Dagster schema — it just stops mutating shared identities to do it.

Worth noting the teardown already refused to delete the `dagsterbootstraps` **CRD** for precisely this reason ("may hold real instances this test knows nothing about"). The RGD delete was the inconsistency, and it's the more dangerous half — losing the shared HelmRepository singleton breaks every Dagster release pointing at it. We hit exactly that failure on a shared cluster once and had to restore the Flux resource by hand.

## [P2] Teardown no longer races KRO
Added a bounded RGD-absence wait before the CRD delete, and a bounded CRD-absence check after it (both surface into the aggregated `failures` rather than being swallowed). This is the same discipline the migration fixture already documents.

## [P3] Migration fixture no longer adopts a fixed identity
`RGD_NAME`, `CR_KIND` and `CR_PLURAL` are now run-suffixed, so a run can only ever touch objects it created and the unconditional pre-delete becomes belt-and-braces instead of a landmine.

## Validation
`bun run typecheck` (lib + examples + tests) clean · **5161 unit tests pass / 0 fail** · `bun run lint` clean. I didn't run the live matrix — these are the cluster-gated fixtures themselves, and I only have a shared cluster to hand, which is the situation the P1 fix exists for.

yehudacohen:
You're right on both, and on P1 you're right that my previous fix didn't go far enough — thanks for pushing back. `1349329`.

## [P1] No shared RGD is written at all now
I'd treated `resourceVersion` as sufficient, and it isn't: it's **concurrency** protection, not **ownership** protection. Two concrete consequences I'd underweighted — the bundle carries `allow-breaking-changes`, so replacing `dagster-bootstrap` migrates the CRD real instances depend on; and a full replace discards labels/annotations owned by whatever system manages that object.

So the `'shared'` path is gone entirely — **there is no `replaceClusterCustomObject` or `patchClusterCustomObject` left in the fixture**. `applyRgd` now:

- rewrites **both** `metadata.name` and `spec.schema.kind` with a run-unique suffix (kind matters too — it's what KRO derives the CRD plural from, so leaving it shared would still collide), and returns the plurals KRO will register so the assertions follow the unique kinds;
- **fails closed**: if a name it's about to create already exists it aborts rather than adopting, and an API error while checking is *not* treated as "free".

The live coverage is unchanged in substance — same generated schema, same `object` typings and preserve-unknown flags, same KRO admission path — just under a throwaway kind nothing depends on. I also made the probe identity run-unique; it was fixed-name, carrying the same concurrency exposure as the migration fixture.

Verified structurally rather than by reading the diff: the only cluster-object mutations remaining are one `create` and the teardown `delete`, both on `${name}-${RUN_ID}`.

## [P2] Absence proven by 404 only
The `.catch(() => false)` inferred absence from *any* error, so a transient API failure read as "gone" — cleanup would delete the CRD while the RGD still existed, KRO re-registers it, and the test reports success having leaked the CRD. Both fixtures now share an `isNotFound` helper (404 only), treat a **timeout as an explicit failure** instead of falling through, and **verify CRD absence** after deleting it. Since both kinds are run-unique, teardown can remove both CRDs without touching a shared definition.

Worth noting I'd propagated the same defect into the `waitForRgdAbsent` helper I added last round — so the fix is applied in both files, not just the one you flagged.

## Validation
`bun run typecheck` (lib + examples + tests) clean · **5161 unit tests pass / 0 fail** · lint clean.

Still not run: the live matrix. These *are* the cluster-gated fixtures, and the only cluster I have is shared — which is the exact scenario the P1 fix exists for. They need a run against a disposable cluster before this merges; that's the one gap I can't close from here.

yehudacohen:
Ran the live matrix on a disposable orbstack cluster (KRO **v0.9.2**, `CELOmitFunction=true` — the production-pinned version). It found **two bugs in my own fix**, now corrected in `ed79446`.

**Final result: 4 pass / 0 fail.**

## What the live run caught

1. **Double-suffixed kind.** The probe composition already declares run-unique identities (`PROBE_RGD`/`PROBE_KIND`, which the CR assertions reference), so `applyRgd`'s unconditional suffix produced `ObjectSchemaProbe<id><id>` and the API rejected the CR outright:
   ```
   kind: Invalid value: "ObjectSchemaProbe<id>": must be ObjectSchemaProbe<id><id>
   ```
   Suffixing is now idempotent.

2. **Hand-built CRD plural.** KRO derives the plural as `kind.toLowerCase() + 's'`. I wrote `schemamigrationprobes<id>`, putting the `s` *before* the run suffix, while KRO registered `schemamigrationprobe<id>s`. Every CRD lookup missed, the fixture timed out waiting for a schema registered under the other name — and cleanup then deleted a name that didn't exist, **leaking the real CRD**. `CR_PLURAL` is now derived from `CR_KIND`.

Neither was reachable offline: typecheck and 5161 unit tests passed over both. KRO's plural derivation and the apiserver's kind-match validation are only observable against a real cluster.

## Teardown verified clean
After the passing run the cluster has **no** probe RGDs, **no** probe CRDs and **no** test namespaces. Pleasingly, the one artifact left behind was the *earlier buggy run's* leaked CRD — which is direct evidence for fix 2, and exactly the failure mode P2 was about. I removed it and left the cluster pristine.

## Where this leaves the three original findings
- **P1** — no `replace`/`patch` calls remain; the only cluster mutations are one `create` and the teardown `delete`, both on run-unique names, with a fail-closed guard. Confirmed against a real cluster.
- **P2** — absence proven by 404 only, timeouts are explicit failures, CRD absence verified. Confirmed: teardown now leaves nothing.
- **P3** — run-unique identity, now correctly derived.

All that plus typecheck (lib + examples + tests) and lint clean.

yehudacohen:
Both correct, and finding 1 also exposed a hole in my *verification*, not just my code. Fixed in `8036b65`.

## [P2] Every registered CRD is now retained and deleted
You were right, and the leak was worse than stated — **both** earlier live runs leaked it, including the one I reported as clean:

```
dagsterhelmrepositorypy3ozns.kro.run   ← run 1
dagsterhelmrepositoryqqhvn1s.kro.run   ← run 2 ("passing")
```

`dagsterbootstrapqqhvn1s` was absent, confirming your diagnosis exactly: only the bootstrap plural was retained. Teardown now deletes **every** plural the run registered.

The reason I missed it: my cleanup check grepped for `objectschemaprobe|schemamigrationprobe` — the patterns I *expected* — rather than enumerating what was actually there. It's now a full before/after diff of all `kro.run` CRDs, which also catches the inverse (destroying something pre-existing).

## [P2] CR `gone()` uses the same 404 classification
Correct — I'd fixed the RGD and CRD waits and left the CR callback on `.catch(() => true)`. Third instance of the same defect, now using `isNotFound()`.

## A third defect the live run surfaced
Chasing these, the migration fixture failed again — and the cause was that I was **deriving** the plural as `kind.toLowerCase() + 's'`. KRO pluralises with real English rules:

```
RUN_ID 'xrjqu1' → schemamigrationprobexrjqu1s     (+s)
RUN_ID 'a32l9o' → schemamigrationprobea32l9oes    (o → oes, potato→potatoes)
```

So with a random suffix the fixtures passed or failed **depending on its last character**, and the mismatch leaked the very CRD teardown then couldn't find. Both fixtures now **discover** the plural by matching `spec.names.kind` on the registered CRD, which is authoritative. That also let me drop the migration fixture's fixed-name pre-delete — unnecessary once the identity is run-unique, and it was the destructive-adoption vector from the original P3.

## Verification (orbstack, KRO v0.9.2, `CELOmitFunction=true`)
```
4 pass / 0 fail
LEAKED (after − before):  NONE
REMOVED (pre-existing):   NONE
RGDs left: 0     test namespaces left: 0
```
Plus typecheck (lib + examples + tests) and lint clean.

Thanks for the persistence on this one — three rounds, and each round the live cluster found something the offline gates couldn't.

yehudacohen:
Switched both cleanup paths to the shared helper — `17845f0`.

## The fix
Both fixtures now delete generated CRDs via `deleteGeneratedCrdAndWait()` from the harness, which owns the `customresourcecleanup.apiextensions.k8s.io` recovery and is fail-closed: it clears the known finalizer only after listing the kind and proving zero instances remain.

That's exactly why the discovered **kind** is now retained alongside each plural, as you suggested — `GeneratedKind { kind, plural }` in values-admission, `CR_KIND` in schema-migration. My hand-rolled `waitForCrdAbsent` helpers are deleted, so **no raw `deleteCustomResourceDefinition` call remains in either fixture** (verified by grep, not just by reading the diff).

## Live verification (orbstack, KRO v0.9.2, `CELOmitFunction=true`)
```
4 pass / 0 fail
LEAKED (after − before):   NONE
REMOVED (pre-existing):    NONE
Terminating CRDs:          NONE
RGDs left: 0    test namespaces left: 0
```

## One caveat I want to be explicit about
**The finalizer-recovery branch did not engage on my cluster** — 0 "Recovering test-owned CRD" warnings; the CRDs deleted gracefully inside the 30s window. So I have not reproduced the stall you hit, and I can't claim to have exercised that specific code path. What I can say is that the raw path which *cannot* survive a stall is gone, and the recovery is now delegated to the harness helper that other suites already depend on.

Given your environment does reproduce it, a rerun of the two suites on your side is the verification that actually closes this one. Happy to iterate if the hooks still fail.

Also worth noting for the record: this is the fourth round, and the live cluster has caught something on every one — including two bugs introduced by my own earlier fixes. The offline gates (typecheck, 5159 unit tests, lint, CI) have been green throughout and had no purchase on any of it.

PR #164, opened 2026-08-26
fix(clickstack): infer namespace ownership from namespace presence

yehudacohen:
## Summary
- omit the namespace field to use and own the default clickstack Namespace
- pass the namespace field to install into an externally owned destination
- preserve this lifecycle behavior in both direct and KRO factory modes
- document the convention and cover owned/external and namespace-hoisting paths

## Why
The namespace value and its lifecycle authority should not require two parallel knobs. Applik8s already owns its application namespace, so passing that namespace to ClickStack must not create a second owner.

The implementation follows TypeKro integration guidance: ordinary TypeScript branches express the condition and TypeKro differential analysis preserves it for KRO, rather than embedding explicit CEL in the integration.

## Verification
- focused ClickStack and namespace-hoisting tests: 28 passed
- full unit suite through the commit hook: 5,319 passed, 13 skipped, 1 todo, 0 failed
- full typecheck passed before the semantic revision; the revised files are covered by the full pre-commit suite

### DIFF
--- a/docs/api/clickhouse/index.md
+++ b/docs/api/clickhouse/index.md
@@ -144,6 +144,30 @@ User names become CHI configuration path fragments (`<name>/password_sha256_hex`
 
 In `makeClickHouseCluster`, declared user names become **literal keys** in the generated runtime spec schema — `spec.users.signoz.passwordSha256Hex` — so each declared user's credential is a required, typed, ref-safe field.
 
+For credentials generated or managed by another controller, select the Secret-backed
+runtime shape when constructing the cluster. The CHI contains only the Secret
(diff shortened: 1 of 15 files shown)


yehudacohen:
Superseded by merged PR #165 and released in TypeKro v0.33.9. #165 carries the external-namespace lifecycle, Secret-backed ClickHouse/ClickStack credentials, and nested-output lowering from this draft, together with the later credential, readiness, timeout, teardown, and live-test hardening. Rebasing this older draft would reintroduce competing implementations, so closing rather than making the obsolete branch artificially green.

PR #165, opened 2026-08-28
fix(clickstack): activate authenticated OTLP ingestion

yehudacohen:
## Summary

- support Secret-backed ClickHouse and ClickStack credentials without serializing values into RGDs or inline Helm values
- allow a parent deployment graph to own the ClickStack namespace
- upgrade the official ClickStack chart to 3.2.0, create the framework-owned ingestion Team, and attach the authenticated OTLP receiver to all signal pipelines
- make ClickStack readiness include completion of the ingestion bootstrap

## Root cause

The released 0.33.8 ClickStack graph reports its HelmRelease Ready while chart 3.0.1's OpAMP configuration leaves `otlp/hyperdx` detached from every pipeline. The collector health port is ready, but 4317/4318 never bind, so application telemetry gets `ECONNREFUSED`.

## Verification

- 127 focused ClickHouse, ClickStack, dependency-resolution, and CEL-resolution tests passed
- library, examples, and test typechecks passed
- Biome lint passed across 992 files
- production build passed
- Applik8s' OrbStack receipt independently reproduced the released failure with a Ready collector, valid Service/endpoints, and no 4318 listener; the patched end-to-end rerun is being completed against this branch

### DIFF
--- a/docs/api/clickstack/index.md
+++ b/docs/api/clickstack/index.md
@@ -7,15 +7,38 @@ OFFICIAL `clickstack` Helm chart (3.0.x, MIT) from
 **external ClickHouse** you already run (e.g. an Altinity-operator-managed
 [`ClickHouseInstallation`](../clickhouse/)).
 
-## Secrets Caveat
+## Credential modes
 
-`clickhouse.password`, `clickhouse.appPassword`, and `apiKey` travel as **plaintext** runtime spec
-values all the way into the generated HelmRelease's `spec.values.hyperdx.secrets.*` — a Kubernetes
-object stored in etcd, readable by anyone with read access to the HelmRelease/RGD instance
-(`kubectl get helmrelease -o yaml`). This is unlike `clickstackK8sTelemetry`'s `apiKeySecret` (a
-`secretKeyRef` env var — the value never appears in any CR spec). There is currently **no
-existing-Secret alternative** for these three fields. Do not treat this family as production-ready
-for credentials that need stronger-than-etcd-RBAC protection until that gap is closed.
+The backwards-compatible default accepts `clickhouse.password`, `clickhouse.appPassword`, and
+`apiKey` inline. Those values enter the HelmRelease and RGD instance, so use that mode only when
+etcd/RBAC protection is sufficient.
+
+For production, construct the Secret-backed variant:
+
+```typescript
+const clickstack = makeClickstackBootstrap({
+  credentials: { source: 'secretValues' },
+});
+
+await clickstack.factory('kro', { namespace: 'typekro-system' }).deploy({
+  name: 'clickstack',
+  namespace: 'observability',
+  clickhouse: {
+    host: 'clickhouse-observability.observability.svc.cluster.local',
+    username: 'otelcollector',
+  },
+  credentialsSecret: {
+    name: 'clickstack-credentials',
+    valuesKey: 'values.yaml',
+  },
+});
+```
+
+The Secret key is a Helm values fragment containing `hyperdx.secrets` and, when a preconfigured UI
+connection is desired, `hyperdx.deployment.defaultConnections`. Flux merges it before TypeKro's
+non-sensitive inline values. TypeKro deliberately omits those credential-bearing paths from the
+HelmRelease and rejects inline password/API-key fields in this variant. The Secret must be in the
+ClickStack workload namespace because Flux values references are namespace-local.
 
 ## Import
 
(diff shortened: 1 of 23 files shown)


yehudacohen:
OrbStack qualification now passes against head a0b03909.

Verified through the real Applik8s/Alchemy consumer path:
- external application-owned Namespace (no duplicate TypeKro Namespace owner)
- ClickStack chart 3.2.0 and Team bootstrap readiness
- semantic-plan nested composition outputs hydrate from flattened HelmRelease resources
- authenticated OTLP logs, traces, and metrics ingest
- a second emitter after restart is queryable in ClickHouse
- redaction canary remains absent
- managed Alchemy/TypeKro destroy completes and the workload Namespace is absent

Focused TypeKro coverage: 146 passed. Typechecks, lint, build, and diff checks passed. The full pre-commit suite reached 5,330 passing tests twice; each run hit a different existing timing-sensitive timeout, and both timed-out tests passed in isolation.

test/integration/clickstack/bootstrap-composition.test.ts
```diff
@@ -107,6 +149,7 @@ describeOrSkip('ClickStack Bootstrap Composition Integration Tests', () => {
         '../../../src/core/kubernetes/index.js'
       );
       const customApi = createBunCompatibleCustomObjectsApi(kubeConfig);
+      await resetClickstackKroDefinition(kubeConfig);
```
yehudacohen, line 156: **[P1] The required live KRO path does not complete at this head**

I ran this suite against OrbStack with `local-path`. Direct mode passed. The KRO phase created and readied the `clickstack-helm-repository` singleton, then spent more than 18 minutes without creating the consumer `clickstack-bootstrap` RGD or CR; the configured 900-second factory timeout did not terminate it, so I had to interrupt the suite. Please reproduce and diagnose whether this reset/migration path or the singleton-to-consumer artifact-bundle transition is blocking, and make the timeout bound the phase. The KRO integration gate must finish successfully before release.

docs/api/clickstack/index.md
```diff
@@ -2,20 +2,52 @@
...
+  },
+});
+```
+
+The Secret key is a Helm values fragment containing `hyperdx.secrets` and, when a preconfigured UI
+connection is desired, `hyperdx.deployment.defaultConnections`. Flux merges it before TypeKro's
+non-sensitive inline values. TypeKro deliberately omits those credential-bearing paths from the
+HelmRelease and rejects inline password/API-key fields in this variant. The Secret must be in the
+ClickStack workload namespace because Flux values references are namespace-local. The fragment must
+override `hyperdx.secrets.HYPERDX_API_KEY`: an idempotent reconciliation CronJob refuses the chart's
```
yehudacohen, line 42: **[P3] Update the status contract for the new convergence gate**

The later Status Contract section still says `ready`/`phase` derive only from `clickstackHelmRelease.status.conditions`. They now also require the current Team-bootstrap CronJob schedule to have succeeded. Please update that section so users understand why an otherwise-ready HelmRelease can leave the stack in `Installing`.

yehudacohen:
Addressed the review findings at `b60d9097`.

- KRO `deploy()` now enforces one operation-wide timeout across singleton-owner convergence, RGD/instance creation, and readiness, preserving the typed `DeploymentTimeoutError` contract.
- Added a regression that proves the timeout aborts a blocked artifact-bundle apply.
- Corrected the ClickStack live teardown to delete only the composition-owned workload Namespace through `factory.deleteInstance()`; the separate CR/control Namespace remains harness-owned.
- Updated the status documentation to describe generation-aware Helm readiness plus the authoritative Team CronJob generation/success contract.

Verification:
- 5,336 unit tests passed
- focused ClickStack/core tests passed
- typecheck and lint passed
- full direct + KRO OrbStack ClickStack suite passed 6/6 with complete teardown and no leaked test namespaces

CI is running on the pushed head.

test/core/kro-artifact-bundle-standalone.test.ts
```diff
@@ -27,6 +28,65 @@ afterEach(() => {
 });
 
 describe('standalone KRO artifact-bundle execution', () => {
+  it('applies one operation-wide timeout across singleton and root bundle phases', async () => {
```
yehudacohen, line 31: [P2] Exercise the singleton-to-root deadline in this regression test

This test proves that the public KRO deploy timeout aborts a hanging root deployment, but the composition does not declare a singleton and only the first `deployResource()` call is exercised. The live failure this repairs occurred after the singleton owner became ready and before the consumer RGD/CR appeared, so the transition named by the test is still not covered by deterministic CI. Please make this composition singleton-backed, let the singleton phase consume part of the deadline, and then hang the root phase; assert that the root receives the same operation signal and only the remaining deadline rather than a fresh timeout. The OrbStack suite now passes this path, but it is not part of ordinary CI.

yehudacohen:
Addressed in f0f67473. The regression now constructs a real singleton owner, proves the singleton RGD and owner instance complete, then stalls the root RGD. It asserts all three deployments receive the same operation signal and that the original 500 ms deadline aborts the root with DeploymentTimeoutError. Verification: focused KRO/ClickStack tests passed, all TypeScript typechecks passed, Biome/diff checks passed, and the full 5,336-test unit suite passed with zero failures.

